PR #4, opened 2022-01-31
Model start

loneil:
# Description
Init first version of DB model

## Types of changes

## Checklist

## Further comments

Run `npm run migrate` to build tables.
Moved seeding specific oidc providers to a seed, if running in a bcgov oidc context can do `npm run seed`


app/src/db/migrations/20220130133615_000-init.js
```diff
@@ -0,0 +1,129 @@
...
+          username: 'System',
+          active: false
+        },
+      ];
+      return knex('oidc_user').insert(items);
+    })
+    .then(() => knex.schema.createTable('identity_provider', table => {
+      table.string('code').primary();
+      table.string('display').notNullable();
+      table.string('idpAlias');
```
TimCsaky, line 33: kind of wondering if we need this field. i guess it can't hurt and could be useful

app/src/db/migrations/20220130133615_000-init.js
```diff
@@ -0,0 +1,129 @@
...
+
+
+    // Add the rest of the tables
+    .then(() => knex.schema.createTable('permission', table => {
+      table.string('code').primary();
+      table.string('display').notNullable();
+      table.boolean('active').notNullable().defaultTo(true);
+      stamps(knex, table);
+    }))
+    .then(() => knex.schema.createTable('object', table => {
```
TimCsaky, line 50: i guess using the term 'object' is the best choice.. hopefully it doesnt get confused with object as a js construct.
maybe 'file' is clearer.. i'm good with object.. just putting it out there

app/src/db/migrations/20220130133615_000-init.js
```diff
@@ -0,0 +1,129 @@
...
+const uuid = require('uuid');
+
+const CREATED_BY = uuid.NIL;
+
+exports.up = function (knex) {
+  return Promise.resolve()
+
+    // Create user table and add a dummy system user (to link migration CreatedBy stamps)
+    .then(() => knex.schema.createTable('oidc_user', table => {
+      table.string('oidcId').primary();
```
TimCsaky, line 11: do we want to use oidc in the label.. would we ever want to authenticate using tokens from another source (eg webade).
probably not, right

app/src/db/migrations/20220130133615_000-init.js
```diff
@@ -0,0 +1,129 @@
...
+      table.string('display').notNullable();
+      table.boolean('active').notNullable().defaultTo(true);
+      stamps(knex, table);
+    }))
+    .then(() => knex.schema.createTable('object', table => {
+      table.uuid('id').primary();
+      table.string('originalName', 1024).notNullable();
+      table.string('path', 1024).notNullable();
+      table.string('mimeType').notNullable();
+      table.string('uploaderOidcId').references('oidcId').inTable('oidc_user');
```
TimCsaky, line 55: this is good for auditting. 
I guess an object will always be linked to a user.

TimCsaky (approved):
-

app/src/db/migrations/20220130133615_000-init.js
loneil: Probably not for us, but would support the ability to have an alias be different accross envs (like happened to us on CHEFS).
So decouples the lookup code (primary key) with the changeable alias in the provider, but likely always the same.

app/src/db/migrations/20220130133615_000-init.js
```diff
@@ -64,6 +64,97 @@ exports.up = function (knex) {
...
+            OLD.id,
+            SESSION_USER,
+            NEW."updatedBy",
+            now(),
+            'U',
+            v_old_data);
+            RETURN NEW;
+        elsif (TG_OP = 'DELETE') then
+            v_old_data := row_to_json(OLD);
+            insert into public.form_submission_audit ("objectPermissionId", "dbUser", "actionTimestamp", "action", "originalData")
```
TimCsaky, line 139: should be insert into table 'audit_object_permission'

app/src/db/migrations/20220130133615_000-init.js
```diff
@@ -64,6 +64,97 @@ exports.up = function (knex) {
...
+            OLD.id,
+            SESSION_USER,
+            NEW."updatedBy",
+            now(),
+            'U',
+            v_old_data);
+            RETURN NEW;
+        elsif (TG_OP = 'DELETE') then
+            v_old_data := row_to_json(OLD);
+            insert into public.form_submission_audit ("objectId", "dbUser", "actionTimestamp", "action", "originalData")
```
TimCsaky, line 111: should be insert into table 'audit_object'

loneil: Oh nice catch. Thanks!

app/src/db/migrations/20220130133615_000-init.js
loneil: Oh nice catch. Thanks!

jujaga (approved):

TimCsaky (approved):

PR #9, opened 2022-02-07
Updating API Spec

MichaelChung123:
# Description
Creating foundation yaml file for the current architecture of the backend.

## Types of changes
Updating v1.api-spec.yaml to represent the work in progress API in the repo.

## Checklist

- [X] I have read the [CONTRIBUTING](CONTRIBUTING.md) doc
- [X] I have checked that unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -17,30 +17,241 @@ security:
   - BearerAuth: []
     OpenID: []
 paths:
-  /hello:
+  '/object':
     get:
-      summary: Returns hello world result
-      operationId: getHello
+      summary: List all user accessible objects
+      description: >-
+        This endpoint returns a list of all user accessible objects.
+      operationId: GetAccessibleObjects
       tags:
-        - Hello
+        - Object
       responses:
         '200':
-          description: Hello result
+          description: Returns the object information
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/HelloDetail'
+                $ref: '#/components/schemas/ObjectDetails'
         '401':
           $ref: '#/components/responses/UnauthorizedError'
         '403':
           $ref: '#/components/responses/Forbidden'
-        default:
-          description: Unexpected error
+    post:
+      summary: Create an Object
+      description: >-
+        This endpoint will create an object.
+      operationId: PostCreateObject
+      tags:
+        - Object
+      responses:
+        '201':
+          description: Returns an array of the created object id
+          content:
+            application/json:
+                type: array
+                items:
+                  $ref: '#/components/schemas/ObjectDetails'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+  '/object/{objId}':
+    get:
+      summary: Details of a single Object
+      description: >-
+        This endpoint returns all of the information related to
+        objId.
+      operationId: GetObject
+      tags:
+        - Object
+      parameters:
+        - $ref: '#/components/parameters/PathObjectId'
+      responses:
+        '200':
+          description: Returns the object information
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Error'
+                $ref: '#/components/schemas/ObjectDetails'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+    post:
+      summary: Updates an Object
+      description: >-
+        This endpoint will update an object.
+      operationId: PostUpdateObject
+      tags:
+        - Object
+      parameters:
+        - $ref: '#/components/parameters/PathObjectId'
+      responses:
+        '201':
+          description: Returns udpate Object response.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MergeResponse'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+    delete:
+      summary: Deletes the object
+      description: >-
+        This endpoint will delete the Object specific to the objId
+      operationId: DeleteObject
+      tags:
+        - Object
+      parameters:
+        - $ref: '#/components/parameters/PathObjectId'
+      responses:
+        '202':
+          description: Request is acknowledged but not necessarily completed
+          headers:
+            Content-Location:
+              example: /api/v1/status/00000000-0000-0000-0000-000000000000
+              schema:
+                type: string
+                format: uri
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '409':
+          $ref: '#/components/responses/Conflict'
+  '/object/{objId}/versions':
+    get:
+      summary: Returns the object version history
+      description: >-
+        This endpoint returns an array of the information related to
+        the object's version history.
+      operationId: GetListObjectVersion
+      tags:
+        - Versions
+      parameters:
+        - $ref: '#/components/parameters/PathObjectId'
+      responses:
+        '200':
+          description: Returns an array of versions for a specific object
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/ObjectDetails'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+  '/object/{objId}/public':
+    get:
+      summary: Sets an object public property
+      description: >-
+        This endpoint sets the public property for an Object
+      operationId: PatchObjectPublic
+      tags:
+        - Permissions
+      parameters:
+        - $ref: '#/components/parameters/PathObjectId'
+      responses:
+        '200':
+          description: Returns the object information
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ObjectDetails'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+  '/object/{objId}/permissions':
+    get:
+      summary: Returns the object permissions
+      description: >-
+        This endpoint returns an array of the Object's permissions
+      operationId: GetObjectPermissions
+      tags:
+        - Permissions
+      parameters:
+        - $ref: '#/components/parameters/PathObjectId'
+      responses:
+        '200':
+          description: Returns the object information
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ObjectDetails'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+  '/object/{objId}/permissions/{userId}':
+    post:
+      summary: Grants object permissions to a specific user
+      description: >-
+        This endpoint will update the permissions of a specific user
+      operationId: PostUpdateUserPermissions
+      tags:
+        - Permissions
+      parameters:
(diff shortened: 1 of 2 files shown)


app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+              schema:
+                $ref: '#/components/schemas/Error'
+  '/{objId}/public':
+    get:
+      summary: Sets an object public property
+      description: >-
+        This endpoint sets the public property for an Object
+      operationId: PatchObjectPublic
+      tags:
+        - Public
```
TimCsaky, line 219: wondering if this should be in with the 'Permissions' tag/section

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+        This endpoint returns a list of all user accessible objects.
+      operationId: GetAccessibleObjects
+      tags:
+        - Object
+      responses:
+        '200':
+          description: Returns the object information
+          content:
+            application/json:
+              schema:
```
TimCsaky, line 33: this is still WIP.. but i think it will retunr an array so maybe change to:
```                
type: array
items:
  $ref: '#/components/schemas/ObjectDetails'
```

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
   - BearerAuth: []
     OpenID: []
 paths:
-  /hello:
+  '/':
```
jujaga, line 20: All of these paths do not start at the root level - they are at `/object/...` level.

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
     get:
-      summary: Returns hello world result
-      operationId: getHello
+      summary: List all user accessible objects
+      description: >-
+        This endpoint returns a list of all user accessible objects.
+      operationId: GetAccessibleObjects
+      tags:
+        - Object
```
jujaga, line 27: We'll need to talk about how we're tagging all of these endpoints later once we have something more solid as IIRC tags affect how they're clustered/grouped together on ReDoc/Swagger renders.

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+      responses:
+        '200':
+          description: Returns the object information
```
TimCsaky, line 30: 'returns an array of object id's'

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+              schema:
+                $ref: '#/components/schemas/MergeResponse'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '409':
+          $ref: '#/components/responses/Conflict'
```
jujaga, line 70: We don't give a 409 on this endpoint.

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+              schema:
+                $ref: '#/components/schemas/ObjectDetails'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '422':
+          $ref: '#/components/responses/UnprocessableEntity'
```
TimCsaky, line 42: i dont think we need this response

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+              schema:
+                $ref: '#/components/schemas/MergeResponse'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '409':
+          $ref: '#/components/responses/Conflict'
```
jujaga, line 134: We likely won't give a 409 on this endpoint - when you update an object, it'll usually always accept it and just make it a new latest version.

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+          description: Returns create Object response.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MergeResponse'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
```
TimCsaky, line 67: dont think we need a 404

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+                type: string
+                format: uri
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '409':
+          $ref: '#/components/responses/Conflict'
```
jujaga, line 171: Current deletion behavior is a 202 when it gets deleted - repeated calls will yield a 404 since it's no longer there. 409 won't be a planned thing.

app/src/docs/v1.api-spec.yaml
```diff
@@ -83,10 +415,14 @@ components:
...
-      description: Hello content
-      example: Hello World!
+    ObjectDetails:
+      type: object
+      properties:
+        objId:
+          type: string
+          description: This Object instance uuid
+          format: uuid
+          example: 00000000-0000-0000-0000-000000000000
```
jujaga, line 425: Let's attempt to keep these object definitions in alphabetical order on the yaml manifest.

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
                 $ref: '#/components/schemas/Error'
+    post:
+      summary: Updates an Object
+      description: >-
+        This endpoint will update an object.
+      operationId: PostUpdateObject
+      tags:
+        - Object
+      parameters:
+        - $ref: '#/components/parameters/PathMessageId'
```
TimCsaky, line 119: you'll have to rename 'PathObjectId' i think some of these schema names are specific to CHES concepts

jujaga (commented):
Looks reasonably good - we'll be coming back to this later again once things get more solidified for sure. Please take a look at some of the comments and modify as necessary.

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+                $ref: '#/components/schemas/Error'
+    delete:
+      summary: Deletes the object
+      description: >-
+        This endpoint will delete the Object specific to the objId
+      operationId: DeleteObject
+      tags:
+        - Object
+      parameters:
+        - $ref: '#/components/parameters/PathMessageId'
```
TimCsaky, line 151: pathObjectId

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+      responses:
+        '202':
+          description: Request is acknowledged but not necessarily completed
+          headers:
+            Content-Location:
+              description: >-
```
TimCsaky, line 157: not sure if we need anything like this

app/src/docs/v1.api-spec.yaml
jujaga: I think the only ones that will likely come back as an array are the following endpoints:
`POST /object`
`GET /object/{objId}/versions`
The rest of them are likely going to be top level objects instead of arrays - still tentative though.

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+      description: >-
+        This endpoint will delete the permission of an Object with a specific to the objId
+      operationId: DeleteObjectPermissions
+      tags:
+        - Permissions
+      parameters:
+        - $ref: '#/components/parameters/PathMessageId'
+        - $ref: '#/components/parameters/PathUserId'
+      responses:
+        '201':
```
TimCsaky, line 320: i think this should be a 204, but is a 200 in code

app/src/docs/v1.api-spec.yaml
```diff
@@ -17,30 +17,348 @@ security:
...
+                $ref: '#/components/schemas/MergeResponse'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '409':
+          $ref: '#/components/responses/Conflict'
+        '422':
```
TimCsaky, line 334: maybe check the chefs api spec that contains similar operations.
i dont think we would have a 422 or 409 in this api.
at this point anyway

TimCsaky:
we need to update line 5 to 18 of the spec file

TimCsaky:
looking at the other specs we have, it appears we used the singular for the tags. eg: object, version, permission

app/src/docs/v1.api-spec.yaml
MichaelChung123: https://github.com/bcgov/common-object-management-service/pull/9/files#r803069575

According to Jeremy's comment I will be leaving this as is and be changing these endpoints accordingly.
POST /object
GET /object/{objId}/versions

app/src/docs/v1.api-spec.yaml
MichaelChung123: Changed Description and type for these endpoints

app/src/docs/v1.api-spec.yaml
MichaelChung123: removing this response for all GET requests

app/src/docs/v1.api-spec.yaml
MichaelChung123: removing all 404 and 422 responses

app/src/docs/v1.api-spec.yaml
MichaelChung123: Removing this response

app/src/docs/v1.api-spec.yaml
MichaelChung123: Renaming all instances of this to PathObjectId

app/src/docs/v1.api-spec.yaml
MichaelChung123: Removing this response

app/src/docs/v1.api-spec.yaml
MichaelChung123: removed

app/src/docs/v1.api-spec.yaml
MichaelChung123: Should I be removing the 404 and 409 calls then from here?

app/src/docs/v1.api-spec.yaml
MichaelChung123: removing both 409 and 422 responses

app/src/docs/v1.api-spec.yaml
jujaga: Just remove the 409 - we'll still have 202s and 404s on this endpoint I think.

jujaga (approved):
Tentatively approving - structure is good enough for first pass. We will revisit again soon as more of the API solidifies.

TimCsaky (approved):

PR #14, opened 2022-02-23
WIP: Middleware tests

loneil:
# Description
Made a start, but need to discuss what to do about config mocking. Both for different values per test, and also to handle local overrides (they take over from test.json it seems).

## Checklist


jujaga (approved):
Will need to revisit later but looks good for now.

PR #17, opened 2022-03-01
Unit tests for User service

TimCsaky:
# Description

## Checklist

### DIFF
--- a/app/tests/unit/services/user.spec.js
+++ b/app/tests/unit/services/user.spec.js
@@ -36,7 +36,7 @@ beforeEach(() => {
 
 describe('_tokenToUser', () => {
 
-  it('Transforms JWT payload contents into a User Model object', async () => {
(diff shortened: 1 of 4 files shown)


app/tests/unit/services/user.spec.js
```diff
@@ -0,0 +1,254 @@
+const { MockModel, MockTransaction } = require('../../common/dbHelper');
+
+jest.mock('../../../src/db/models/tables/user', () => MockModel);
+jest.mock('../../../src/db/models/tables/identityProvider', () => MockModel);
+
+const service = require('../../../src/services/user');
+
+const oidcId = 'cc8c64b7-a457-456e-ade0-09ff7ee75a2b';
```
jujaga, line 8: Let's try not to use identifiable ids as mock information - instead I suggest using the uuid.NIL value as it's still a valid oidcId and suffices for tests.

jujaga (approved):

PR #37, opened 2022-05-19
Versions

TimCsaky:
# Description

- sql migration for version table, 
  - move orginalName and mimeType columns and data from object to version table
from object table
- object controller updated to manage versions
- version service
- check if versioning enabled on bucket
- api spec minor updates

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -75,7 +75,6 @@ const controller = {
         });
       });
       bb.on('close', async () => {
-
         await Promise.all(objects.map(async (object) => {
           // wait for object and permission db update
           object.dbResponse = await object.dbResponse;
@@ -84,18 +83,15 @@ const controller = {
           // add VersionId to data for the file. If versioning not enabled on bucket. VersionId is undefined
           object.data.VersionId = object.s3Response.VersionId;
         }));
-
         // create version in DB
         const objectVersionArray = objects.map((object) => object.data);
         await versionService.create(objectVersionArray, userId);
-
         // merge returned responses into a result
         const result = objects.map((object) => ({
           ...object.data,
           ...object.dbResponse,
           ...object.s3Response
         }));
-
         res.status(201).json(result);
       });
 
@@ -116,33 +112,27 @@ const controller = {
   async deleteObject(req, res, next) {
     try {
       const objId = addDashesToUuid(req.params.objId);
-      const versionId = req.query.versionId;
-      const filePath = getPath(objId);
+      const data = {
+        filePath: getPath(objId),
+        versionId: req.query.versionId
+      };
       const userId = getCurrentSubject(req.currentUser);
 
-      // if request is to delete a version
-      if (versionId) {
-        // delete version on S3
-        const s3Response = await storageService.deleteObject({ filePath, versionId });
-        const objectVersionId = s3Response.VersionId;
+      // delete version on S3
+      const s3Response = await storageService.deleteObject(data);
 
+      // if request is to delete a version
+      if (data.versionId) {
+        const objectVersionId = s3Response.VersionId;
         // delete version in DB
         await versionService.delete(objId, objectVersionId);
-
         // if other versions in DB, delete object record
         // TODO: synch with versions in S3
         const remainingVersions = await versionService.list(objId);
-        console.log('remainingVersions:', remainingVersions, remainingVersions.length);
-
         if (remainingVersions.length === 0) await objectService.delete(objId);
-
-        res.status(200).json(s3Response);
       }
-
       // else deleting the object
       else {
-        // delete version on S3
-        const s3Response = await storageService.deleteObject({ filePath });
         // if versioning enabled s3Response will contain DeleteMarker: true
         if (s3Response.DeleteMarker) {
           // create DeleteMarker version in DB
@@ -160,8 +150,8 @@ const controller = {
           // delete object record from DB
           await objectService.delete(objId);
         }
-        res.status(200).json(s3Response);
       }
+      res.status(200).json(s3Response);
     } catch (e) {
       next(errorToProblem(SERVICE, e));
     }
@@ -182,9 +172,7 @@ const controller = {
         filePath: getPath(objId),
         versionId: req.query.versionId ? req.query.versionId.toString() : undefined
       };
-
       const response = await storageService.headObject(data);
-
       // Set Headers
       // TODO: Consider adding 'x-coms-public' and 'x-coms-path' headers into API spec?
       controller._setS3Headers(response, res);
@@ -273,6 +261,7 @@ const controller = {
     // TODO: Handle no database scenarios via S3 ListObjectsCommand?
     // TODO: Consider metadata/tagging query parameter design here?
     // TODO: Consider support for filtering by set of permissions?
+    // TODO: handle additional parameters. Eg: deleteMarker, latest
     try {
       const objIds = mixedQueryToArray(req.query.objId);
       const params = {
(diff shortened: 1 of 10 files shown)


app/src/services/version.js
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * @function createVersionOfObjects
+   * Saves a version of each object in an array
+   * @param {object[]} data array of objects each with an `objectId` and version data
+   * @param {object} [etrx=undefined] An optional Objection Transaction object
+   * @returns {Promise<object>} The result of calling the create() method for all objects
+   * @throws The error encountered upon db transaction failure
+   */
+  createVersionOfObjects: async (data = [], userId, etrx = undefined) => {
```
jujaga, line 52: I would suggest renaming this to `createManyObjects` as the focus of the multiplicity is the fact that there are many objects that will need their own versions.

app/src/services/version.js
```diff
@@ -0,0 +1,99 @@
...
+    let trx;
+    try {
+      trx = etrx ? etrx : await Version.startTransaction();
+
+      // wait for all create version calls
+      const response =  Promise.all(data.map(async (object) => {
+        await service.create(object, userId, trx);
+      }));
```
jujaga, line 60: I recommend changing this to use the insert([ array ]).returns("*") pattern instead as this would reduce the volume of explicit SQL calls done within the transaction. For example, if you were uploading 3 files, using the current promise/map paradigm would yield 3 separate insert statements, whereas doing it with the insert array instead will only yield one insert statement. The latter would be more efficient in both network traffic as well as sql query operation.

app/src/services/version.js
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * @function delete
+   * Delete a version record of an object
+   * @param {string} objId The object uuid
+   * @param {string} versionId The version ID
+   * @param {object} [etrx=undefined] An optional Objection Transaction object
+   * @returns {Promise<object>} The result of running the delete operation
+   * @throws The error encountered upon db transaction failure
+   */
+  delete: async ({objId, versionId = null}, etrx = undefined) => {
```
jujaga, line 79: We can default the function signature to explicit 3 arguments instead of casting within an object.
```js
delete: async (objId, versionId = undefined, etrx = undefined) => {
```
versionId should probably default to undefined instead as that'll be easier to cascade.

app/src/middleware/authorization.js
```diff
@@ -54,6 +54,7 @@ const currentObject = async (req, _res, next) => {
     }
   } catch (err) {
     log.warn(err.message, { function: 'currentObject' });
+    return new Problem(404, { detail: `Object ${req.params.objId} not found` }).send(_res);
```
TimCsaky, line 57: need to discuss with team if we want to return 403 or 404

app/src/db/migrations/20220516000000_001-version.js
```diff
@@ -0,0 +1,73 @@
...
+exports.up = function (knex) {
+  return Promise.resolve()
+    // create version table
+    .then(() => knex.schema.createTable('version', table => {
+      table.uuid('id').primary();
+      table.string('versionId', 1024);
+      table.uuid('objectId').references('id').inTable('object').notNullable().onUpdate('CASCADE').onDelete('CASCADE');
+      table.string('originalName', 255);
+      table.string('mimeType', 255);
+      stamps(knex, table);
```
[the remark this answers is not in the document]
loneil: Yeah I agree, if not needed would probably be good not to have to track and up date when a new version is added? We know latest from the s3 response if I recall.

app/src/middleware/authorization.js
loneil: 403 if I'm understanding right. Since it would throw if you provide a object ID and then that isn't found in the DB, so we wouldn't want to expose the not found as a 404 (also other errors could happen that don't necessarily specify a Not Found)

app/src/services/version.js
```diff
@@ -0,0 +1,168 @@
...
+      if (!etrx) await trx.commit();
+      return response;
+    } catch (err) {
+      if (!etrx && trx) await trx.rollback();
+      throw err;
+    }
+  },
+
+  /**
+   * @function createVersionOfObjects
```
loneil, line 53: function name mismatch

app/src/services/version.js
```diff
@@ -0,0 +1,168 @@
...
+   * @function makeLatest
+   * Updates most recent or provided version with isLatest = true and all other versions of object with isLatest = false
+   * @param {string} objId The object uuid
+   * @param {string} versionId The version ID
+   * @param {string} uuid of the current user
+   * @param {object} [etrx=undefined] An optional Objection Transaction object
+   * @returns {Promise<object>} The result of running the update operation
+   * @throws The error encountered upon db transaction failure
+   */
+  makeLatest: async (objId, versionId = undefined, currentUserId = SYSTEM_USER, etrx = undefined) => {
```
loneil, line 126: This part I'm not sure we need? Tracking the latest in the database... worth a discussion

app/src/middleware/authorization.js
TimCsaky: i'm now thinking we just keep the blanket 403 for now

app/src/db/migrations/20220516000000_001-version.js
```diff
@@ -0,0 +1,73 @@
...
+    .then((rows) => {
+      const versions = rows.map(o => ({
+        id: uuidv4(),
+        objectId: o.id,
+        originalName: o.originalName,
+        mimeType: o.mimeType,
+        createdBy: SYSTEM_USER
+      }));
+
+      return versions.length ? knex('version').insert(versions) : Promise.resolve();
```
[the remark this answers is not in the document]
jujaga: Undefined behavior with suggestion; keeping as is

app/src/db/migrations/20220516000000_001-version.js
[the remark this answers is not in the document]
jujaga: Discussed - going to not track isLatest for now; most temporal information can be derived from ordering on CreatedAt

app/src/services/version.js
TimCsaky: IsLatest is a S3 construct to mark a version as 'current'. 
For example, to check if a object has been soft-deleted, you need to check if a delete-marker is the latest version.
Jeremy suggested using the createdAt field to check which one is latest instead.
I'll try removing this isLatest stuff and seeing how it goes.

app/src/services/version.js
TimCsaky: now that i dont have the isLatest stuff happening, there's no need to call create() for each object. Done multi-row insert as suggested here.

app/src/services/version.js
```diff
@@ -93,14 +60,12 @@ const service = {
         .where('versionId', versionId)
         .throwIfNotFound();
 
-      // if other versions exist, make next most recent version current
+      // get number of other remaining versions
       const versionCount = await Version.query(trx)
         .count('id as CNT')
         .where({ objectId: objId });
       const count = parseInt(versionCount[0].CNT);
```
jujaga, line 67: Since we only care about the number of remaining rows that match a criteria, why aren't we just using the `resultSize` method instead?
Ref: https://vincit.github.io/objection.js/api/query-builder/other-methods.html#resultsize
```js
const count = await Version.query(trx).resultSize();
```
On a second note, do we even need to get a count for this? We're not using this value anywhere afterwards.

app/src/controllers/object.js
```diff
@@ -100,15 +112,45 @@ const controller = {
...
-        storageService.deleteObject(data) // Attempt deletion operation
-      ]);
-      res.status(200).json(results[1]);
+      const versionId = req.query.versionId;
+      const filePath = getPath(objId);
+      const userId = getCurrentSubject(req.currentUser);
+      // if request is to delete a version
+      if (versionId) {
+        // delete version on S3
+        const s3Response = await storageService.deleteObject({ filePath, versionId });
```
TimCsaky, line 121: const s3Response = await storageService.deleteObject({ filePath, versionId ? versionId : undefined });

jujaga (approved):
Fantastic work 👍

kamorel (approved):

PR #38, opened 2022-05-27
Schema validation middleware

kamorel:
# Description

Implements [express-validation](https://www.npmjs.com/package/express-validation) middleware to validate incoming requests against a defined schema. Any requests not matching will automatically be rejected with a 422.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2650

## Types of changes

## Checklist


app/app.js
```diff
@@ -117,12 +118,16 @@ apiRouter.use('/v1', v1Router);
...
 // eslint-disable-next-line no-unused-vars
 app.use((err, _req, res, _next) => {
   if (err instanceof Problem) {
     err.send(res);
-  } else {
+  }
+  else if (err instanceof ValidationError) {
+    return res.status(err.statusCode).json(err);
+  }
+  else {
```
jujaga, line 130: Style recommendation
```js
  } else if (err instanceof ValidationError) {
    return res.status(err.statusCode).json(err);
  } else {
```

app/tests/unit/routes/v1/user.spec.js
```diff
@@ -0,0 +1,97 @@
...
+const router = require('../../../../src/routes/v1/user');
+
+const { userController } = require('../../../../src/controllers');
+
+// Express Server
+const basePath = '/api/v1/user';
+const app = expressHelper(basePath, router);
+
+// Mock config library - @see {@link https://stackoverflow.com/a/64819698}
+jest.mock('config');
```
jujaga, line 32: We should have this be declared earlier - probably best to have this be along with all the other jest.mock definitions above?

app/tests/unit/routes/v1/user.spec.js
```diff
@@ -0,0 +1,97 @@
...
+
+describe(`GET ${basePath}`, () => {
+  // TODO: Ensure other middleware funcs are called once
+  const validatorSearchUserSpy = jest.spyOn(validator, 'searchUsers');
+  const ctrlrSearchUserSpy = jest.spyOn(userController, 'searchUsers');
+
+  beforeEach(() => {
+    validatorSearchUserSpy.mockReset();
+    ctrlrSearchUserSpy.mockReset();
```
jujaga, line 41: We may be able to avoid needing to explicitly spy reset declare on all of these if we use `jest.resetAllMocks()` at a higher beforeEach level?
Ref: <https://jestjs.io/docs/jest-object#jestresetallmocks>

app/tests/unit/routes/v1/user.spec.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  it('should call controller with known query params', async () => {
+    // eslint-disable-next-line no-unused-vars
+    validatorSearchUserSpy.mockImplementation((_req, _res, next) => next());
+    // eslint-disable-next-line no-unused-vars
+    ctrlrSearchUserSpy.mockImplementation((req, res, next) => res.status(200).end());
```
jujaga, line 48: Not a particular fan of having to ignore no-unused-vars repeatedly like this. If we have to continue doing this (aka no way to circumvent this due to implementation mock limits...), I'd recommend we add this following line to the very top of this file instead to reduce verbosity:
```js
/* eslint-disable no-unused-vars */
```

app/tests/unit/validator/user.spec.js
```diff
@@ -0,0 +1,314 @@
...
+        expect(Array.isArray(identityId.items)).toBeTruthy();
+        expect(identityId.items).toHaveLength(1);
+        expect(identityId.items).toEqual(expect.arrayContaining([
+          expect.objectContaining({
+            type: 'string',
+            rules: expect.arrayContaining([
+              expect.objectContaining({
+                name: 'guid',
+                args: {
+                  options: { version: 'uuidv4'}
```
jujaga, line 127: Inconsistent spacing - check vscode right click autoformat.

app/src/validator/user.js
```diff
@@ -0,0 +1,36 @@
+const { validate, Joi } = require('express-validation');
+
+const schema = {
+  searchUsers: {
+    query: Joi.object({
+      userId: Joi.array().items(Joi.string().guid({
```
TimCsaky, line 6: We're using the mixedQueryToArray() method to turn one or more values for parameters (eg: userId and identityId) into an array, but they can be provided as a string in the query.
eg: /api/user?userId=098090-324-24-23-4234 should be valid.

app/src/validator/user.js
```diff
@@ -0,0 +1,36 @@
...
+      identityId: Joi.array().items(Joi.string().guid({
+        version: 'uuidv4'
+      })),
+      idp: Joi.array().items(Joi.string()),
+      username: Joi.string(),
+      email: Joi.string().email(),
+      firstName: Joi.string(),
+      fullName: Joi.string(),
+      lastName: Joi.string(),
+      active: Joi.boolean(),
```
TimCsaky, line 18: Jeremy's isTruthy method means any of these could be valid: `'true', 't', 'yes', 'y', '1']`

app/src/validator/user.js
```diff
@@ -0,0 +1,36 @@
...
+      userId: Joi.array().items(Joi.string().guid({
+        version: 'uuidv4'
+      })),
+      identityId: Joi.array().items(Joi.string().guid({
+        version: 'uuidv4'
+      })),
+      idp: Joi.array().items(Joi.string()),
+      username: Joi.string(),
+      email: Joi.string().email(),
+      firstName: Joi.string(),
```
TimCsaky, line 15: wonder if we need to use `Joi.string().alphanum();` to be extra strict

app/tests/unit/routes/v1/user.spec.js
```diff
@@ -0,0 +1,97 @@
...
+
+    const response = await request(app).get(`${basePath}`).query({
+      userId: ['11bf5b37-e0b8-42e0-8dcf-dc8c4aefc000'],
+      idp: ['IDIR'],
+      active: 'true'
+    });
```
TimCsaky, line 54: minor point: we could check that this test object is being passed to the validator or controller. i think that's done with: `expect(validatorSearchUserSpy).toHaveBeenCalledWith();`
also we could consider defining this dummy object at a higher scope further up the file to be re-used.

app/src/validators/user.js
```diff
@@ -0,0 +1,75 @@
...
+      ),
+      idp: Joi.alternatives().try(
+        Joi.array().items(Joi.string().max(255)),
+        Joi.string().max(255)
+      ),
+      username: Joi.string().pattern(/^[\w\-\s]+$/).max(255),
+      email: Joi.string().max(255).email(),
+      firstName: Joi.string().pattern(/^[\w\-\s]+$/).max(255),
+      fullName: Joi.string().pattern(/^[\w\-\s]+$/).max(255),
+      lastName: Joi.string().pattern(/^[\w\-\s]+$/).max(255),
```
jujaga, line 30: We should have `username`, `firstName` and `lastName` use the `string.alphaNum()` restriction; the only one that should have the regex is fullName as we need to allow alphanumeric strings with one or more spaces.

app/src/routes/v1/permission.js
```diff
@@ -1,5 +1,6 @@
 const routes = require('express').Router();
 
+const { permissionValidator } = require('../../validators');
```
jujaga, line 3: Style: This should be introduced after the permissionController - have some degree of alphabetical ordering based on require path sorting.

app/src/validators/permission.js
```diff
@@ -0,0 +1,55 @@
...
+
+  addPermissions: {
+    params: Joi.object({
+      objId: uuidv4
+    }),
+    body: Joi.array().items(
+      Joi.object().keys({
+        userId: Joi.string().guid({
+          version: 'uuidv4'
+        }).required(),
```
jujaga, line 29:
```js
userId: uuidv4.required(),
```

app/src/validators/permission.js
```diff
@@ -0,0 +1,55 @@
...
+        permCode: Joi.string().max(255).required(),
+      })
+    ).required(),
+  },
+
+  removePermissions: {
+    params: Joi.object({
+      objId: uuidv4
```
jujaga, line 37: Missing required?
```js
objId: uuidv4.required(),
```

app/src/validators/user.js
```diff
@@ -1,5 +1,8 @@
 const { validate, Joi } = require('express-validation');
 
+const truthyModel = Joi.boolean()
+  .truthy('true', 1, '1', 't', 'yes', 'y', 'false', 0, '0', 'f', 'no', 'n');
+
```
jujaga, line 5: This is a good candidate for validators/common.js

app/tests/unit/validators/common.spec.js
```diff
@@ -0,0 +1,177 @@
...
+  it('is a uuidv4', () => {
+    expect(model).toBeTruthy();
+    expect(model.type).toEqual('string');
+    expect(Array.isArray(model.rules)).toBeTruthy();
+    expect(model.rules).toHaveLength(1);
+    expect(model.rules).toEqual(expect.arrayContaining([
+      expect.objectContaining({
+        name: 'guid',
+        args: {
+          options: { version: 'uuidv4'}
```
jujaga, line 18: VSCode formatter

app/tests/unit/validators/common.spec.js
```diff
@@ -0,0 +1,177 @@
...
+
+  it('rejects the schema with incorrect guid', () => {
+    expect('notauuidv4').not.toMatchSchema(uuidv4);
+  });
+});
+
+describe('uuidv4MultiModel', () => {
+  const model = uuidv4MultiModel.describe();
+
+  it('is an alternatives', () => {
```
jujaga, line 36: 'is of type alternatives' is clearer at a glance.

app/tests/unit/validators/common.spec.js
```diff
@@ -0,0 +1,177 @@
...
+
+  it('rejects the schema with incorrect guid', () => {
+    expect('notauuidv4').not.toMatchSchema(uuidv4MultiModel);
+  });
+});
+
+describe('stringMultiModel', () => {
+  const model = stringMultiModel.describe();
+
+  it('is an alternatives', () => {
```
jujaga, line 109: 'is of type alternatives' is clearer at a glance.

app/tests/unit/validators/permission.spec.js
```diff
@@ -0,0 +1,186 @@
...
+      });
+    });
+
+    describe('permCode', () => {
+      const permCode = query.keys.permCode;
+
+      it('is the expected schema', () => {
+        expect(permCode).toEqual(stringMultiModel.describe());
+      });
+    });
```
jujaga, line 46: I can't quite remember, but permCode may be one of the string values we explicitly define as an enum on the API spec. If that is the case, we could go one step further and do string matches to ensure that only values that are within that enum set pass.

app/tests/unit/validators/permission.spec.js
```diff
@@ -0,0 +1,186 @@
...
```
jujaga, line 77: I can't quite remember, but permCode may be one of the string values we explicitly define as an enum on the API spec. If that is the case, we could go one step further and do string matches to ensure that only values that are within that enum set pass.

app/tests/unit/validators/permission.spec.js
```diff
@@ -0,0 +1,186 @@
...
```
jujaga, line 184: I can't quite remember, but permCode may be one of the string values we explicitly define as an enum on the API spec. If that is the case, we could go one step further and do string matches to ensure that only values that are within that enum set pass.

app/src/validators/permission.js
```diff
@@ -0,0 +1,60 @@
...
+const { uuidv4, uuidv4MultiModel, permCodeMultiModel } = require('./common');
+const { Permissions } = require('../components/constants');
+
+const schema = {
+  searchPermissions: {
+    query: Joi.object({
+      userId: uuidv4MultiModel,
+      objId: uuidv4MultiModel,
+      permCode: permCodeMultiModel
+    }).min(1)
```
TimCsaky, line 11: Was this min(1) intentional?
providing no query parameters does return a result from the API (all the records in the object_permissions table)
then again maybe it's nice to make the search more prescriptive, i dont see anyone needing the whole list for all users and objects..

app/src/validators/permission.js
```diff
@@ -0,0 +1,60 @@
...
+    query: Joi.object({
+      userId: uuidv4MultiModel,
+      objId: uuidv4MultiModel,
+      permCode: permCodeMultiModel
+    })
+  },
+
+  addPermissions: {
+    params: Joi.object({
+      objId: uuidv4
```
TimCsaky, line 24: not sure .. do you need to mark this as required?

app/app.js
```diff
@@ -117,11 +118,13 @@ apiRouter.use('/v1', v1Router);
...
 app.use(/(\/api)?/, apiRouter);
 
-// Handle 500
+// Handle ValidationError & 500
 // eslint-disable-next-line no-unused-vars
 app.use((err, _req, res, _next) => {
   if (err instanceof Problem) {
     err.send(res);
+  } else if (err instanceof ValidationError) {
+    return res.status(err.statusCode).json(err);
```
TimCsaky, line 127: it may be a good idea to log validation errors. it has value from a security perspective to monitor brute force attacks.
something from the STRA checklist.

app/src/validators/permission.js
```diff
@@ -0,0 +1,60 @@
...
+      })
+    ).required(),
+  },
+
+  removePermissions: {
+    params: Joi.object({
+      objId: uuidv4
+    }),
+    query: Joi.object({
+      userId: uuidv4MultiModel,
```
TimCsaky, line 45: a userId and permCode is also required according to the code in the removePermissions() controller.

TimCsaky (changes requested):
looking good. i like the way youve organised this. found a few minor details that looked suspicious

app/app.js
jujaga: Not necessarily. Validation errors arise from external entity network traffic that contains parameters or queries that do not match our specifications. If we do decide to log these kinds of errors, they would need to be flagged at either the verbose or the lower debug level. Most enterprise level applications do not log anything above and beyond the http level for most situations. You can catch brute force attacks with just http request layer stuff already without any need for more explicit logging.

app/src/validators/common.js
```diff
@@ -0,0 +1,40 @@
...
+    Permissions.DELETE,
+    Permissions.MANAGE
+  )),
+  Joi.string().max(255).valid(
+    Permissions.CREATE,
+    Permissions.READ,
+    Permissions.UPDATE,
+    Permissions.DELETE,
+    Permissions.MANAGE
+  )
```
jujaga, line 37: You could do something clever with asserting permissions with an array spread operator like the following:
```js
Joi.array().items(Joi.string().max(255).valid(...Permissions),
Joi.string().max(255).valid(...Permissions)
```
Since we already define Permissions object, and we're effectively using all of the object, we can just let JS handle the de-structuring assignment. This has the added benefit of allowing us to need to refactor less in the event we add a new permissions constant, as it'd just immediately propagate here with spread syntax instead of needing us to explicitly name it.

app/src/validators/permission.js
```diff
@@ -0,0 +1,60 @@
...
+    body: Joi.array().items(
+      Joi.object().keys({
+        userId: uuidv4.required(),
+        permCode: Joi.string().max(255).required().valid(
+          Permissions.CREATE,
+          Permissions.READ,
+          Permissions.UPDATE,
+          Permissions.DELETE,
+          Permissions.MANAGE
+        ),
```
jujaga, line 35: Refer to destructuring comment.

app/tests/unit/validators/common.spec.js
```diff
@@ -0,0 +1,328 @@
+const crypto = require('crypto');
+const jestJoi = require('jest-joi');
+expect.extend(jestJoi.matchers);
+
+const { Permissions } = require('../../../src/components/constants');
+const { alphanumModel, truthyModel, uuidv4, uuidv4MultiModel, stringMultiModel, permCodeMultiModel } = require('../../../src/validators/common');
```
jujaga, line 6: With this many discrete attributes being imported, may be good to multi-line them for readability purposes (we're well past the 80 col character limit here)

jujaga (commented):
Consider the destructuring proposal; it'll resolve some of Code Climate's duplicate code complaints. Starting to come together. 👍

app/src/validators/permission.js
kamorel: Jeremy and I discussed marking the params required, but as a param of the route they are already explicitly required to call it. It would also require a duplicate object of uuidv4 with the required tag to be created or the unit testing will fail. So we scrapped the idea.

jujaga (approved):

TimCsaky (approved):
looks good to me

PR #44, opened 2022-06-23
Version-related API structure and basic unit tests

TimCsaky:
# Description

ticket:[ https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2648]( https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2648)

* Edited version related parts of API spec

* Added unit tests for 
  * the deleteObject method in the object controller dealing with versions.
  * getBucketVersioning method in storage service

* tested in different auth and db modes and without versioning enabled

## Types of changes

Documentation (non-breaking change with enhancements to documentation)

## Checklist

## Further comments

i wonder if we should remove the object.active field from db responses. (maybe override the model somehow)
and `active=true|false` parameter for the object/search endpoint in the API spec (and  for now (keep it in code) but not advertise it. You cant update that field in the db, its always true. it might confuse people, especially with the delete/soft-delete stuff. when we figure out what we can use active for, add it back in.

### DIFF
--- a/app/tests/unit/controllers/object.spec.js
+++ b/app/tests/unit/controllers/object.spec.js
@@ -10,27 +10,23 @@ const mockResponse = () => {
   res.json = jest.fn().mockReturnValue(res);
   return res;
 };
-
 // Mock config library - @see {@link https://stackoverflow.com/a/64819698}
 jest.mock('config');
 
-const storageDeleteObjectSpy = jest.spyOn(storageService, 'deleteObject');
-const objectDeleteSpy = jest.spyOn(objectService, 'delete');
-const versionCreateSpy = jest.spyOn(versionService, 'create');
-const versionDeleteSpy = jest.spyOn(versionService, 'delete');
-const versionListSpy = jest.spyOn(versionService, 'list');
-
-
 const res = mockResponse();
 
-
 describe('deleteObject', () => {
+  // mock service calls
+  const storageDeleteObjectSpy = jest.spyOn(storageService, 'deleteObject');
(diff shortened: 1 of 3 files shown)


app/tests/unit/controllers/object.spec.js
```diff
@@ -0,0 +1,112 @@
...
+jest.mock('config');
+
+const storageDeleteObjectSpy = jest.spyOn(storageService, 'deleteObject');
+const objectDeleteSpy = jest.spyOn(objectService, 'delete');
+const versionCreateSpy = jest.spyOn(versionService, 'create');
+const versionDeleteSpy = jest.spyOn(versionService, 'delete');
+const versionListSpy = jest.spyOn(versionService, 'list');
+
+
+const res = mockResponse();
```
jujaga, line 24: In general, it's best practice to not declare spies and mock constructs as top-level objects in a Jest test; values in general should only be declared at the most specific block scope possible.

These values should be declared under the `describe('deleteObject', () => {` block scope. (Reinsert to Line 28)

Also when using spyOn, you generally want an associated `beforeEach()` lifecycle that will `mockReset()` them in order to guarantee test atomicity. Without this, you may be testing behavior with unintentional side effects.

jujaga (commented):
Make sure your tests are atomic. You can do this by randomly changing the ordering of the it... blocks and then re-running your tests -> they should all be able to pass, regardless of the order they are defined.

kamorel (approved):

PR #46, opened 2022-06-30
Add endpoints for add/replace/delete metadata

kamorel:
# Description

Adds new endpoints to add, replace, or delete metadata from an object. Each call will create a copy (and thus new version) of the object with the changes applied.

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -1,7 +1,7 @@
 const busboy = require('busboy');
 const { v4: uuidv4, NIL: SYSTEM_USER } = require('uuid');
 
-const { AuthMode } = require('../components/constants');
+const { AuthMode, S3 } = require('../components/constants');
 const errorToProblem = require('../components/errorToProblem');
 const {
   addDashesToUuid,
@@ -57,19 +57,16 @@ const controller = {
       const objPath = getPath(objId)
 
       const latest = await storageService.headObject({ filePath: objPath });
-      if (latest.ContentLength > 5368709120) {
+      if (latest.ContentLength > S3.MAXCOPYOBJECTLENGTH) {
         throw new Error('Cannot copy an object larger than 5GB');
       }
       else {
-        if (Object.keys(req.query).length <= 0 ||
-          Object.keys(req.query).length === 1 && req.query.hasOwnProperty('versionId')) {
+        if (!Object.keys(req.query).length || Object.keys(req.query).length && req.query.versionId) {
           // 422 when no keys given
           res.status(422).end();
         }
         else {
-          const metadataToAppend = req.query;
-
-          if (metadataToAppend.hasOwnProperty('versionId')) delete metadataToAppend['versionId'];
+          const { versionId, ...metadataToAppend } = req.query;
 
           const data = {
             copySource: objPath,
@@ -77,11 +74,10 @@ const controller = {
             metadata: {
               ...latest.Metadata,  // Take existing metadata first
               ...metadataToAppend, // Append new metadata
-              name: latest.Metadata.name,  // Always enforce name and id key behavior
-              id: latest.Metadata.id
+              id: latest.Metadata.id // Always enforce id key behavior
             },
             metadataDirective: 'REPLACE',
-            versionId: req.query.versionId ? req.query.versionId.toString() : undefined
+            versionId: versionId ? versionId.toString() : undefined
           };
 
           const response = await storageService.copyObject(data);
@@ -314,24 +310,22 @@ const controller = {
       const objPath = getPath(objId)
 
       const latest = await storageService.headObject({ filePath: objPath });
-      if (latest.ContentLength > 5368709120) {
+      if (latest.ContentLength > S3.MAXCOPYOBJECTLENGTH) {
         throw new Error('Cannot copy an object larger than 5GB');
       }
       else {
-        const keysToRemove = mixedQueryToArray(req.query.key);
-        let metadata = latest.Metadata;
+        const { versionId, ...newMetadata } = req.query;
+        const keysToRemove = mixedQueryToArray(newMetadata.key);
 
-        if (keysToRemove.hasOwnProperty('versionId')) delete keysToRemove['versionId'];
+        let metadata = undefined;
 
-        if (keysToRemove !== undefined && keysToRemove.length > 0) {
-          keysToRemove.forEach(x => {
-            if (metadata.hasOwnProperty(x)) delete metadata[x];
-          });
-        } else {
-          metadata = undefined;
+        // Generate object subset by subtracting/omitting defined keys via filter/inclusion
+        if (keysToRemove && keysToRemove.length) {
+          metadata = Object.fromEntries(
+            Object.entries(latest.Metadata)
+              .filter(([key]) => !keysToRemove.includes(key))
+          )
         }
-
-        // TODO: if keys is empty remove all but default
 
         const data = {
           copySource: objPath,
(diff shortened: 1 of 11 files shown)


app/src/controllers/object.js
```diff
@@ -43,6 +43,57 @@ const controller = {
...
+      else {
+        if (Object.keys(req.query).length <= 0 ||
+          Object.keys(req.query).length === 1 && req.query.hasOwnProperty('versionId')) {
+          // 422 when no keys given
+          res.status(422).end();
+        }
+        else {
+          const metadataToAppend = req.query;
+
+          if (metadataToAppend.hasOwnProperty('versionId')) delete metadataToAppend['versionId'];
```
jujaga, line 72: It's bad practice to modify the request object stream like this. It's always better to instead construct a new object and copy out the subset of items you need from there instead. The way it's implemented right now, `metadataToAppend` is just a reference to the same `req.query` object, so if you deleted an attribute off from there, you're actually deleting it for both. Instead, you can achieve the same logic by using object destructuring to construct a new object.
```js
const { versionId, ...metadataToAppend } = req.query;
```

app/src/controllers/object.js
```diff
@@ -43,6 +43,57 @@ const controller = {
...
+
+          const data = {
+            copySource: objPath,
+            filePath: objPath,
+            metadata: {
+              ...latest.Metadata,  // Take existing metadata first
+              ...metadataToAppend, // Append new metadata
+              name: latest.Metadata.name,  // Always enforce name and id key behavior
```
jujaga, line 80: If we do this, we will be disallowing users to "rename their file and apply as a new version". I think we should allow users to redefine their filename `name` if they really wanted to, so this line may not be necessary (already covered by ...latest.Metadata).

app/src/controllers/object.js
```diff
@@ -43,6 +43,57 @@ const controller = {
...
+              id: latest.Metadata.id
+            },
+            metadataDirective: 'REPLACE',
+            versionId: req.query.versionId ? req.query.versionId.toString() : undefined
```
jujaga, line 84: If you use the object destructuring comment from earlier, this could be changed to something simpler:
```js
versionId: versionId ? versionId.toString() : undefined
```

app/src/controllers/object.js
```diff
@@ -249,6 +300,109 @@ const controller = {
...
+      if (latest.ContentLength > 5368709120) {
+        throw new Error('Cannot copy an object larger than 5GB');
+      }
+      else {
+        const keysToRemove = mixedQueryToArray(req.query.key);
+        let metadata = latest.Metadata;
+
+        if (keysToRemove.hasOwnProperty('versionId')) delete keysToRemove['versionId'];
+
+        if (keysToRemove !== undefined && keysToRemove.length > 0) {
```
jujaga, line 326:
```js
if (!keysToRemove && keysToRemove.length) {
```

app/src/controllers/object.js
```diff
@@ -249,6 +300,109 @@ const controller = {
...
+          keysToRemove.forEach(x => {
+            if (metadata.hasOwnProperty(x)) delete metadata[x];
```
jujaga, line 328: You may be able to simplify the logic here by doing it decoratively instead of imperatively using ES2019 style code:
```js
// Generate object subset by subtracting/omitting defined keys via filter/inclusion
const newMetadataSet = Object.fromEntries(
  Object.entries(metadata)
    .filter(([key]) => !keysToRemove.includes(key))
)
```
As a side note, this syntax is all supported in Node.js 12.x and above, so we're good.
Ref: https://stackoverflow.com/a/56592365

jujaga (commented):
A ton of `missing semicolon` and `Do not access Object.prototype method 'hasOwnProperty' from target object` warnings. For the latter, you can always just attempt to access the property directly, and then evaluate the truthiness of the value to see if it exists or not; don't need to invoke the hasOwnProperty method at all.

TimCsaky (commented):
looks good. Good that you did the controller and service as well for these!
i guess it would be better to add the tag related routes in another PR. And those controllers will work a bit differently.

not sure what we decided about the existing POST /object/:objId (to update an object) and whether you were able to modify metadata there as well. I think we discussed this: https://discord.com/channels/689896523848613952/689896524406194236/991494695609974855 probably could go in a separate PR. I think we should merge this soon..  unit tests could go in a separate PR.

would need some kind

app/src/docs/v1.api-spec.yaml
```diff
@@ -344,6 +344,68 @@ paths:
...
+        - $ref: "#/components/parameters/Path-ObjectId"
+        - $ref: "#/components/parameters/Query-VersionId"
+      responses:
+        "204": 
+          description: The resource was deleted successfully.
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        default:
```
jujaga, line 406: If we're going to be doing 422 rejections, that will also need to be defined here as potential HTTP response type.

app/src/docs/v1.api-spec.yaml
```diff
@@ -344,6 +344,68 @@ paths:
...
+      summary: Replaces metadata of an object
+      description: >-
+        Creates a copy and new version of the object with the given metadata replacing the existing. Multiple Key/Value pairs can be provided in the header for the metadata.
+      operationId: replaceMetadata
+      tags:
+        - Object
+      parameters:
+        - $ref: "#/components/parameters/Path-ObjectId"
+        - $ref: "#/components/parameters/Query-VersionId"
+        - $ref: "#/components/parameters/Header-Metadata"
```
jujaga, line 379: Header references should come first (alphabetical ordering as well)

app/src/docs/v1.api-spec.yaml
```diff
@@ -344,6 +344,68 @@ paths:
...
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        default:
+          $ref: "#/components/responses/Error"
+    delete:
+      summary: Delete metadata of an object.
+      description: >-
+        Creates a copy and new version of the object with the given metadata removed. Multiple Key/Value pairs can be provided in the header for the metadata. If no metadata headers are given then all metadata will be removed.
```
jujaga, line 392: Break up this into a multiline description. It's too long.

app/src/services/object.js
```diff
@@ -85,6 +85,7 @@ const service = {
    * @param {string} [params.originalName] Optional filename string to match on
    * @param {string} [params.path] Optional canonical S3 path string to match on
    * @param {string} [params.mimeType] Optional mimeType string to match on
+   * @param {string} [params.metadata] Optional metadata string to match on
```
jujaga, line 88: If we're not implementing the search side of things just yet, let's not add this in just yet to the JSDocs.

app/src/services/object.js
```diff
@@ -99,6 +100,7 @@ const service = {
       .modify('filterUserId', params.userId)
       .modify('filterOriginalName', params.originalName)
       .modify('filterMimeType', params.mimeType);
+    // .modify('filterMetadata', params.metadata);
```
jujaga, line 103: If we're not implementing the search side of things just yet, let's not add this in as a comment yet.

app/src/controllers/object.js
```diff
@@ -43,6 +44,54 @@ const controller = {
...
+        else {
+          const { versionId } = req.query;
+
+          const data = {
+            copySource: objPath,
+            filePath: objPath,
+            metadata: {
+              ...latest.Metadata,  // Take existing metadata first
+              ...metadataToAppend, // Append new metadata
+              id: latest.Metadata.id // Always enforce id key behavior
```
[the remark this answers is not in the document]
kamorel: This will be done in the validation layer via Joi

app/tests/unit/controllers/object.spec.js
```diff
@@ -8,13 +8,159 @@ const mockResponse = () => {
...
+  const next = jest.fn();
+
+  // response from S3
+  const GoodResponse = {
+    ContentLength: 1234,
+    Metadata: { id: 1, foo: 'bar' }
+  };
+  const BadResponse = {
+    ContentLength: MAXCOPYOBJECTLENGTH + 1
+  };
```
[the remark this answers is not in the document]
kamorel: Can't change these as this is what the head request is returning.

TimCsaky (approved):

PR #47, opened 2022-07-04
Metadata and Tag database migration and models

TimCsaky:
# Description

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2640
 
-  database migration of 4 new tables, triggers, etc. spent a while moving existing data around (see file ...002-metadata-tag.js)
- 4 new ORM models.
- moved originalName into metadata.name
- added objectId to metadata.id
- created metadata service > addMetadata method that creates the metadata DB records
- metadata service is currently only invoked with the create and update object operations
- updated the object modifier (used by object search) 'filterByName' to query related metadata table.
We have lots to do with that search query. I suspect we may have to do some more of the query building in the object service rather than doing it all with modifiers.
- updated API spec

## Checklist

### DIFF
--- a/app/src/db/migrations/20220627000000_002-metadata-tags.js
+++ b/app/src/db/migrations/20220627000000_002-metadata-tags.js
@@ -17,7 +17,7 @@ exports.up = function (knex) {
     // create version_metadata table
     .then(() => knex.schema.createTable('version_metadata', table => {
       table.primary(['versionId', 'metadataId']);
-      table.uuid('versionId').notNullable().references('id').inTable('version');
+      table.uuid('versionId').notNullable().references('id').inTable('version').onDelete('CASCADE').onUpdate('CASCADE');
       table.integer('metadataId').notNullable().references('id').inTable('metadata');
       stamps(knex, table);
     }))
@@ -36,7 +36,7 @@ exports.up = function (knex) {
     // create version_tag table
     .then(() => knex.schema.createTable('version_tag', table => {
       table.primary(['versionId', 'tagId']);
-      table.uuid('versionId').notNullable().references('id').inTable('version');
+      table.uuid('versionId').notNullable().references('id').inTable('version').onDelete('CASCADE').onUpdate('CASCADE');
       table.integer('tagId').notNullable().references('id').inTable('tag');
       stamps(knex, table);
     }))
--- a/app/src/db/migrations/20220627000000_002-metadata-tags.js
+++ b/app/src/db/migrations/20220627000000_002-metadata-tags.js
@@ -50,10 +50,10 @@ exports.up = function (knex) {
     AFTER UPDATE OR DELETE ON version_tag
     FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();`))
 
-  /**
-   * for each version, move originalName and objectId to records in metadata table
-   * and create joining version_metadata records
-   */
+    /**
+     * for each version, move originalName and objectId to records in metadata table
+     * and create joining version_metadata records
+     */
     .then(() => knex('version').where('deleteMarker', false))
     .then((rows) => {
       const versions = rows.map(v => ({
--- a/app/src/db/migrations/20220627000000_002-metadata-tags.js
+++ b/app/src/db/migrations/20220627000000_002-metadata-tags.js
@@ -3,24 +3,23 @@ const stamps = require('../stamps');
 exports.up = function (knex) {
   return Promise.resolve()
 
-    // // create metadata table
+    // create metadata table
     .then(() => knex.schema.createTable('metadata', table => {
       table.specificType(
         'id',
         'integer GENERATED ALWAYS AS IDENTITY PRIMARY KEY'
       );
-      table.string('key').index();
-      table.string('value').index();
+      table.string('key', 255).index();
+      table.string('value', 255).index();
       table.unique(['key', 'value']);
-
     }))
 
     // create version_metadata table
     .then(() => knex.schema.createTable('version_metadata', table => {
+      table.primary(['versionId', 'metadataId']);
       table.uuid('versionId').notNullable().references('id').inTable('version');
       table.integer('metadataId').notNullable().references('id').inTable('metadata');
       stamps(knex, table);
-      table.primary(['versionId', 'metadataId']);
     }))
 
     // create tag table
@@ -29,19 +28,18 @@ exports.up = function (knex) {
         'id',
         'integer GENERATED ALWAYS AS IDENTITY PRIMARY KEY'
       );
-      table.string('key').index();
-      table.string('value').index();
+      table.string('key', 255).index();
+      table.string('value', 255).index();
       table.unique(['key', 'value']);
     }))
 
     // create version_tag table
     .then(() => knex.schema.createTable('version_tag', table => {
+      table.primary(['versionId', 'tagId']);
       table.uuid('versionId').notNullable().references('id').inTable('version');
       table.integer('tagId').notNullable().references('id').inTable('tag');
       stamps(knex, table);
-      table.primary(['versionId', 'tagId']);
     }))
-
 
     // Create metadata audit trigger
     .then(() => knex.schema.raw(`CREATE TRIGGER audit_version_metadata_trigger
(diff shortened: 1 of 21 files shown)


app/src/db/migrations/20220627000000_002-metadata-tags.js
```diff
@@ -0,0 +1,122 @@
+const stamps = require('../stamps');
+
+exports.up = function (knex) {
+  return Promise.resolve()
+
+    // // create metadata table
```
jujaga, line 6: Extra `//`

app/src/db/migrations/20220627000000_002-metadata-tags.js
```diff
@@ -0,0 +1,122 @@
...
+    .then(() => knex.schema.createTable('metadata', table => {
+      table.specificType(
+        'id',
+        'integer GENERATED ALWAYS AS IDENTITY PRIMARY KEY'
+      );
+      table.string('key').index();
+      table.string('value').index();
+      table.unique(['key', 'value']);
+
```
jujaga, line 15: Extra whitespace line

app/src/db/migrations/20220627000000_002-metadata-tags.js
```diff
@@ -0,0 +1,122 @@
...
+
+    // Create metadata audit trigger
+    .then(() => knex.schema.raw(`CREATE TRIGGER audit_version_metadata_trigger
+    AFTER UPDATE OR DELETE ON version_metadata
+    FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();`))
+    // Create tag audit trigger
+    .then(() => knex.schema.raw(`CREATE TRIGGER audit_version_tag_trigger
+    AFTER UPDATE OR DELETE ON version_tag
+    FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();`))
```
jujaga, line 53: While technically we aren't updating the explicit metadata/tag table records themselves, they do still need to be audited. Please add those triggers in.

app/src/db/migrations/20220627000000_002-metadata-tags.js
```diff
@@ -0,0 +1,122 @@
...
+          { versionId: row.versionId, metadataId: insertIdArray[1].id }
+        ]);
+      });
+      await Promise.all(promises);
+    })
+
+    // remove column originalName from version table
+    .then(() => knex.schema.alterTable('version', table => {
+      table.dropColumn('originalName');
+    }));
```
jujaga, line 86: I'm not convinced we should remove the originalName column just yet -> likely need further discussion.
If we do go through with removing the originalName column, then there will need to be minimal associated changes to our control layer to make sure that the metadata we're recording is getting written into the new models; if we merged this PR as is now, it would likely fail or error.

app/src/db/migrations/20220627000000_002-metadata-tags.js
```diff
@@ -0,0 +1,122 @@
...
+          .merge()
+          .returning('id');
+        // add joining records
+        return knex('version_metadata').insert([
+          { versionId: row.versionId, metadataId: insertIdArray[0].id },
+          { versionId: row.versionId, metadataId: insertIdArray[1].id }
+        ]);
+      });
+      await Promise.all(promises);
+    })
```
jujaga, line 81: We should never be using async/await syntax at all in migrations. This should be refactored using promise chain style. The following "should" be equivalent if I read through it correctly:
```js
.then((rows) => {
  const versions = rows.map(v => ({
    value: v.originalName,
    versionId: v.id,
    objectId: v.objectId
  }));

  return Promise.all(versions.map((row) => {
    // insert into metadata table
    knex('metadata').insert([
      { key: 'name', value: row.value },
      { key: 'id', value: row.objectId }
    ]).onConflict(['key', 'value'])
      .merge()
      // Return just id column
      .returning('id')
      // add joining records
      .then((result) => {
        return knex('version_metadata').insert(result.map((metadata) => ({
          metadataId: metadata.id,
          versionId: row.versionId
        })))
      })
  }));
})
```
I'm not sure if we should be running `.merge()` on here or if it's better to do a `.ignore()` as in these cases, what we're wanting to put into the metadata already exists and doesn't need to be updated at all.

app/src/db/migrations/20220627000000_002-metadata-tags.js
```diff
@@ -0,0 +1,122 @@
...
+      .where('metadata.key', 'name')
+    )
+    .then(async (versionMetadata) => {
+      const promises = versionMetadata.map((vm) => {
+        return knex('version')
+          .update({ originalName: vm.value })
+          .where({ id: vm.versionId });
+      });
+      await Promise.all(promises);
+    })
```
jujaga, line 111: We should never be using async/await syntax at all in migrations. This should be refactored using promise chain style. The following "should" be equivalent if I read through it correctly:
```js
.then((versionMetadata) => {
  return Promise.all(versionMetadata.map((vm) => {
    return knex('version')
      .update({ originalName: vm.value })
      .where({ id: vm.versionId });
  }));
})
```

app/src/db/migrations/20220627000000_002-metadata-tags.js
```diff
@@ -0,0 +1,122 @@
...
+  return Promise.resolve()
+
+    // // create metadata table
+    .then(() => knex.schema.createTable('metadata', table => {
+      table.specificType(
+        'id',
+        'integer GENERATED ALWAYS AS IDENTITY PRIMARY KEY'
+      );
+      table.string('key').index();
+      table.string('value').index();
```
jujaga, line 13: Even though Knex defaults to 255, we should generally be explicit about length declarations, especially in migration files. This is because in the future, if Knex decides to change its default, we would have just introduced a very hard to track down bug.
```js
table.string('key', 255).index();
table.string('value', 255).index();
```

app/src/db/migrations/20220627000000_002-metadata-tags.js
```diff
@@ -0,0 +1,122 @@
...
+    }))
+
+    // create tag table
+    .then(() => knex.schema.createTable('tag', table => {
+      table.specificType(
+        'id',
+        'integer GENERATED ALWAYS AS IDENTITY PRIMARY KEY'
+      );
+      table.string('key').index();
+      table.string('value').index();
```
jujaga, line 33: Even though Knex defaults to 255, we should generally be explicit about length declarations, especially in migration files. This is because in the future, if Knex decides to change its default, we would have just introduced a very hard to track down bug.
```js
table.string('key', 255).index();
table.string('value', 255).index();
```

app/src/db/migrations/20220627000000_002-metadata-tags.js
TimCsaky: these changes worked. thanks!

app/src/db/migrations/20220627000000_002-metadata-tags.js
TimCsaky: the `if_modified_func()` that is triggered inserts the updated record (as JSON) into the audit log. 
it also looks for the 'updatedBy' field in the changed row to put in the audit record 'updatedByUsername' column.
..so the table being audited needs to have an updatedBy column. it can have a NULL value but must exist.

to audit this table i believe we would either need to modify that function or have the updatedBy/At columns.
probably something we should defer until we have a feature in coms where people can actually update/delete the tags themselves

app/src/services/version.js
```diff
@@ -97,6 +101,7 @@ const service = {
    * Updates a version of an object.
    * Typically happens when updating the 'null-version' created for an object
    * on a non-versioned or version-suspnded bucket.
+   * Replaces metadta that already exists on this version by default
```
jujaga, line 104: Typo on "metadta"

app/src/validators/object.js
```diff
@@ -37,7 +37,7 @@ const schema = {
   searchObjects: {
     query: Joi.object({
       objId: scheme.guid,
-      originalName: type.alphanum,
+      name: type.alphanum,
```
jujaga, line 40: This will fail on "foobar.pdf". We'll need to allow that. For now we'll reduce the specificity to just `Joi.string()`.

jujaga (approved):

kamorel (approved):

PR #50, opened 2022-07-14
Object tagging

kamorel:
# Description

Adds functionality to add tags on object create & update via query parameters.
Adds 3 new endpoints to modify tags on objects (add/replace/delete).

## Types of changes

## Checklist


TimCsaky (changes requested):
just remembered this is in draft state so maybe ignore my comments for now

app/src/controllers/object.js
```diff
@@ -391,6 +470,41 @@ const controller = {
...
   },
 
+  /**
+   * @function replaceTags
+   * Replaces the tag set on the requested object
+   * @param {object} req Express request object
+   * @param {object} res Express response object
+   * @param {function} next The next callback function
+   * @returns {function} Express middleware function
+  */
```
jujaga, line 480: General JSDoc indentation spacing seems a bit off in this file. Double check when you can.

app/src/controllers/object.js
```diff
@@ -91,6 +90,47 @@ const controller = {
...
+    try {
+      const objId = addDashesToUuid(req.params.objId);
+      const objPath = getPath(objId);
+      const { versionId, ...newTags } = req.query;
+      const objectTagging = await storageService.getObjectTagging({ filePath: objPath, versionId });
+
+      // Join new and existing tags then filter duplicates
+      let newSet = Object.entries(newTags).map(([k, v]) => ({ Key: k, Value: v }));
+      if (objectTagging.TagSet) newSet = newSet.concat(objectTagging.TagSet);
+      newSet = newSet.filter((v, i, a) => a.findIndex(v2 => (v2.Key === v.Key)) === i);
```
jujaga, line 111: It may be worthwhile having more descriptive variable names in the anonymous function as they aren't innately clear what they represent without re-reading multiple times. Key/value only is fine making them k and v as that's pretty clear, but here not too sure what "a" is.

app/src/controllers/object.js
```diff
@@ -253,6 +293,45 @@ const controller = {
...
+
+      const data = {
+        filePath: objPath,
+        tags: newTags,
+        versionId: versionId ? versionId.toString() : undefined
+      };
+
+      let response = await storageService.deleteObjectTagging(data);
+      if (newTags) {
+        response = await storageService.putObjectTagging(data);
```
jujaga, line 326: This section seems dangerous as we're potentially overwriting the data coming from deleteObjectTagging. Should it be an either or situation instead?

app/src/routes/v1/storage.js
```diff
@@ -0,0 +1,19 @@
...
+
+/** */
+routes.get('/encryption', requireSomeAuth, (req, res, next) => {
+  storageController.getEncryption(req, res, next);
+});
+
+/** */
+routes.patch('/encryption', requireSomeAuth, (req, res, next) => {
+  storageController.putEncryption(req, res, next);
+});
```
jujaga, line 16: Not completely sure if we should be implementing these storage endpoints yet. IIRC there's a more general getBucketCommand that could also be doing the same operations as this as well; and if it does provides encryption status, we should consider lumping it in there instead.

app/src/controllers/storage.js
```diff
@@ -0,0 +1,29 @@
...
+
+  async putEncryption(req, res, next) {
+    try {
+      const response = await storageService.putBucketEncryption();
+      res.status(200).json(response);
+    } catch (e) {
+      next(errorToProblem(SERVICE, e));
+    }
+  },
+};
```
jujaga, line 27: It's good to keep this code around, but probably scope it out to a different branch of work for now as we still need to get a better sense of where we're headed with encryption processes in general.

app/src/services/storage.js
```diff
@@ -86,6 +92,34 @@ const objectStorageService = {
     return this._s3Client.send(new DeleteObjectCommand(params));
   },
 
+  /**
+   * @function deleteObjectTagging
+   * Deletes the tags of the object at `filePath`
+   * @param {string} options.filePath The filePath of the object
+   * @param {number} [options.versionId] Optional specific versionId for the object
+   */
+  async deleteObjectTagging({ filePath, tags, versionId }) {
```
jujaga, line 101: Tags isn't being used or wired into the function. Was this an oversight? If it is, we should probably set tags to be an empty set of some form for a default.

app/src/services/storage.js
```diff
@@ -86,6 +92,34 @@ const objectStorageService = {
...
+
+  /**
+   * @function getBucketEncryption
+   */
+  async getBucketEncryption() {
+    const params = {
+      Bucket: bucket
+    };
+
+    return await this._s3Client.send(new GetBucketEncryptionCommand(params));
```
jujaga, line 120: We should not be awaiting and force unwrapping this command as the storage service by convention always returns a promise.

app/src/services/storage.js
```diff
@@ -168,6 +232,24 @@ const objectStorageService = {
...
+      ServerSideEncryptionConfiguration: {
+        Rules: [{
+          ApplyServerSideEncryptionByDefault: {
+            SSEAlgorithm: 'AES256'
+          }
+        }]
+      }
+    };
+
+    return await this._s3Client.send(new PutBucketEncryptionCommand(params));
```
jujaga, line 250: We should not be awaiting and force unwrapping this command as the storage service by convention always returns a promise.

app/src/services/storage.js
```diff
@@ -202,6 +283,28 @@ const objectStorageService = {
...
+   * @param {string} options.tags Array of key/value pairs
+   * @param {number} [options.versionId] Optional specific versionId for the object
+   */
+  async putObjectTagging({ filePath, tags, versionId }) {
+    const params = {
+      Bucket: bucket,
+      Key: filePath,
+      Tagging: {
+        TagSet:
+          tags
```
jujaga, line 299: One line this

app/src/controllers/object.js
```diff
@@ -91,6 +90,47 @@ const controller = {
...
+   * @param {object} res Express response object
+   * @param {function} next The next callback function
+   * @returns {function} Express middleware function
+  */
+  async addTags(req, res, next) {
+    try {
+      const objId = addDashesToUuid(req.params.objId);
+      const objPath = getPath(objId);
+      const { versionId, ...newTags } = req.query;
+      const objectTagging = await storageService.getObjectTagging({ filePath: objPath, versionId });
```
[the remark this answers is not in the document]
kamorel: No, tags are either hard delete or replace. You have to get the existing set first in order to append.

app/src/services/storage.js
```diff
@@ -56,13 +61,14 @@ const objectStorageService = {
    * @param {string} [options.versionId=undefined] Optional versionId to copy from
    * @returns {Promise<object>} The response of the delete object operation
    */
-  copyObject({ copySource, filePath, metadata, metadataDirective = MetadataDirective.COPY, versionId = undefined }) {
+  copyObject({ copySource, filePath, metadata, metadataDirective = MetadataDirective.COPY, taggingDirective = TaggingDirective.COPY, versionId = undefined }) {
```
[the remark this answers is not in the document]
kamorel: Tagging directive internally by default is COPY, i'm just explicitly mentioning it here.

app/src/routes/v1/storage.js
```diff
@@ -0,0 +1,19 @@
+const routes = require('express').Router();
+const { storageController } = require('../../controllers');
+const { requireSomeAuth } = require('../../middleware/featureToggle');
+const { checkAppMode } = require('../../middleware/authorization');
+
+routes.use(checkAppMode);
+
+/** */
+routes.get('/encryption', requireSomeAuth, (req, res, next) => {
```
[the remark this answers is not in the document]
kamorel: Removing the storage route and controller for the time being.

app/src/controllers/object.js
```diff
@@ -91,6 +90,47 @@ const controller = {
...
+   * @param {object} req Express request object
+   * @param {object} res Express response object
+   * @param {function} next The next callback function
+   * @returns {function} Express middleware function
+  */
+  async addTags(req, res, next) {
+    try {
+      const objId = addDashesToUuid(req.params.objId);
+      const objPath = getPath(objId);
+      const { versionId, ...newTags } = req.query;
```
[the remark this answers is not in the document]
kamorel: No, it can be anywhere in the query params.

app/src/controllers/object.js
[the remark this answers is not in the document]
kamorel: Restructured. Delete is a hard delete. But as putObjectTagging is a replace by default I only need to call one or the other.

app/tests/unit/services/storage.spec.js
```diff
@@ -39,6 +44,133 @@ describe('_s3Client', () => {
   });
 });
 
+describe.only('copyObject', () => {
```
jujaga, line 47: Need to remove `.only`.

TimCsaky (approved):

jujaga (approved):

PR #57, opened 2022-07-29
Generate internal uuid for COMS user, getCurrentUserId function

TimCsaky:
# Description

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2744

- identify user with db > user.userId filed instead of 'sub' field in JWT
- replace use of getCurrentSubject() with getCurrentUserId

tested in SSO Gold realm.. seems to be working

## Checklist

### DIFF
--- a/app/tests/unit/middleware/authorization.spec.js
+++ b/app/tests/unit/middleware/authorization.spec.js
@@ -3,7 +3,7 @@ const config = require('config');
 const { NIL: SYSTEM_USER } = require('uuid');
 
 const { checkAppMode, currentObject, hasPermission } = require('../../../src/middleware/authorization');
-const { objectService, permissionService, storageService } = require('../../../src/services');
+const { objectService, permissionService, storageService, userService } = require('../../../src/services');
 const { AuthMode, AuthType, Permissions } = require('../../../src/components/constants');
 const utils = require('../../../src/components/utils');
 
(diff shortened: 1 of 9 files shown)


app/tests/unit/middleware/authorization.spec.js
```diff
@@ -166,7 +166,8 @@ describe('hasPermission', () => {
     ])('should call next %i times given hasDb %s and authMode %s', (nextCount, hasDb, mode) => {
       const sendCount = 1 - nextCount;
       getAppAuthModeSpy.mockReturnValue(mode);
-      getCurrentSubjectSpy.mockReturnValue(undefined);
+      //getCurrentSubjectSpy.mockReturnValue(undefined);
```
jujaga, line 169: There is still value in making sure that getCurrentSubjectSpy is tracked - although we're now using it as a nested argument call in getCurrentUserId, we should at minimum still check to make sure we're invoking getCurrentSubjectSpy at least once as an assertion. I suspect you'll want to uncomment these series of mockReturnValues, and also add in something like the following in the appropriate if block:
```js
expect(getCurrentSubjectSpy).toHaveBeenCalledTimes(someappropriatevariable);
```
In general, when we do a unit test, we want to ensure that all function calls that we mock are also asserted to be called within the function as that will tell you when the incoming code changes are no longer invoking an external function/side-effect that it used to depend on (early warning sign of potential functionality divergence).

jujaga (approved):

PR #63, opened 2022-08-17
Implement metadata search functionality

jujaga:
# Description

This PR focuses on exposing and implementing arbitrary metadata search capabilities.

[SHOWCASE-2338](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2338)

## Types of changes

Documentation (non-breaking change with enhancements to documentation)

## Checklist

### DIFF
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -617,16 +617,18 @@ const controller = {
    */
   async searchObjects(req, res, next) {
     // TODO: Handle no database scenarios via S3 ListObjectsCommand?
-    // TODO: Consider metadata/tagging query parameter design here?
+    // TODO: Consider tagging query parameter design here
     // TODO: Consider support for filtering by set of permissions?
     // TODO: handle additional parameters. Eg: deleteMarker, latest
     try {
       const objIds = mixedQueryToArray(req.query.objId);
+      const metadata = getMetadata(req.headers);
       const params = {
         id: objIds ? objIds.map(id => addDashesToUuid(id)) : objIds,
         name: req.query.name,
         path: req.query.path,
         mimeType: req.query.mimeType,
+        metadata: metadata && Object.keys(metadata).length ? metadata : undefined,
         public: isTruthy(req.query.public),
         active: isTruthy(req.query.active)
       };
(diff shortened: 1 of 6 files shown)


kamorel (approved):

app/src/db/models/tables/objectModel.js
```diff
@@ -35,43 +35,69 @@ class ObjectModel extends Timestamps(Model) {
...
+
+        if (name) {
+          subqueries.push(Version.query()
+            .select('version.id')
             .joinRelated('metadata')
-            .where('value', 'ilike', `%${value}%`)
-            .andWhere('key', 'name'));
+            .where('metadata.key', 'name')
+            .where('metadata.value', 'ilike', `%${name}%`));
+        }
```
TimCsaky, line 82: just a recent thought. i guess we could have just removed 'name' as a parameter, and make them look for it using a metadata header. but this is better.
and we could use this pattern if there are other bits of more 'standard' metadata, such as 'retention' or whatever if we added those as separate query parameters in the future.

app/src/controllers/object.js
```diff
@@ -617,16 +617,18 @@ const controller = {
...
     // TODO: handle additional parameters. Eg: deleteMarker, latest
     try {
       const objIds = mixedQueryToArray(req.query.objId);
+      const metadata = getMetadata(req.headers);
       const params = {
         id: objIds ? objIds.map(id => addDashesToUuid(id)) : objIds,
         name: req.query.name,
         path: req.query.path,
         mimeType: req.query.mimeType,
+        metadata: metadata && Object.keys(metadata).length ? metadata : undefined,
```
TimCsaky, line 631: this && stuff was in case they provide a key with no value, right?
by the way, i wonder if we need to urldecode keys and values.

TimCsaky (approved):
glad you figured the queries out, It seems like a good pattern. We can try and do some performance testing at some point. but looks good.

app/src/controllers/object.js
jujaga: At the controller level, we need to check for two things:

1. 1) Is metadata truthy (aka does it exist)
2. 2) Is the object actually populated with "something"

We need the second check because in the event there is nothing/undefined, we want to outright skip that portion of logic in the later query building phase.

Url encoding and decoding for headers should be fine for metadata; I didn't see too many examples in the wild where we had to sanitize metadata headers.

PR #67, opened 2022-08-30
Separate tag service layer

TimCsaky:
# Description

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2777

separate tag service layer into distinct 'atomic' methods for add / update / delete db operations.
previously we were doing 'delete all then insert..' (which creates misleading entries in our audit logs) where we want to only insert new.

## Types of changes

## Checklist

### DIFF
--- a/app/src/components/utils.js
+++ b/app/src/components/utils.js
@@ -248,10 +248,10 @@ const utils = {
    * @function toLowerKeys
    * Converts all key names for all objects in an array to lowercase
    * @param {object[]} arr Array of tag objects (eg: [{Key: k1, Value: V1}])
-   * @returns {object[]} Array of objects (eg: [{key: k1, value: V1}])
+   * @returns {object[]} Array of objects (eg: [{key: k1, value: V1}]) or undefined if empty
    */
   toLowerKeys(arr) {
-    if (!arr) return [];
+    if (!arr) return undefined;
     return arr.map(obj => {
       return Object.fromEntries(
         Object.entries(obj).map(([key, value]) => {
--- a/app/src/components/utils.js
+++ b/app/src/components/utils.js
@@ -22,6 +22,17 @@ const utils = {
   },
 
   /**
+   * @function toLowerCaseKeys
+   * @param {object} obj an object with keys of mixed-case
+   * @returns {object} the input object with keys converted to lowercase
+   */
+  toLowerCaseKeys(obj){
+    return Object.fromEntries(Object.entries(obj).map(entry => {
+      return [entry[0].toLowerCase(), entry[1]];
+    }));
+  },
+
+  /**
    * @function delimit
    * Yields a string `s` that will always have a trailing delimiter. Returns an empty string if falsy.
    * @param {string} s The input string
@@ -182,6 +193,18 @@ const utils = {
    */
   getKeyValue(input) {
     return Object.entries(input).map(([k, v]) => ({ key: k, value: v }));
+  },
+
+  /**
+   * @function getTagsByKeyValue
+   * get tag objects in array that have given key and value
+   * @param {object[]} tags and array of tags (eg: [{ key: 'a', value: '1'}, { key: 'b', value: '1'}]
+   * @param {string} key the string to match in the tag's `key` property
+   * @param {string} value the string to match in the tag's `value` property
+   * @returns {object[]} an array of matching tag objects
+   */
+  getTagsByKeyValue(tags, key, value){
+    return tags.find(tag => (tag.key === key && tag.value === value));
   },
 
   /**
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -13,11 +13,11 @@ const {
   addDashesToUuid,
   getAppAuthMode,
   getKeyValue,
+  toLowerCaseKeys,
   getMetadata,
   getPath,
   isTruthy,
   mixedQueryToArray,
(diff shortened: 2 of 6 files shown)


app/src/services/tag.js
```diff
@@ -7,42 +7,196 @@ const { Tag, VersionTag } = require('../db/models');
...
-          .insert(relateArray);
+        // get id's of all input tags
+        const dbTags = await service.createTags(tags, trx);
+
+        // get all currently associated tags
+        const allArray = await VersionTag.query(trx)
+          .modify('filterVersionId', versionId);
+
+        // associate remaining tags
+        const newJoins = dbTags.filter(({ id }) => {
```
TimCsaky, line 76: turns out we do need to do some more diff stuff, that was missing.

app/src/services/tag.js
```diff
@@ -7,42 +7,196 @@ const { Tag, VersionTag } = require('../db/models');
...
+    let trx;
+    try {
+      trx = etrx ? etrx : await Tag.startTransaction();
+      let response = [];
+
+      // delete joining records
+      response = await VersionTag.query(trx)
+        .allowGraph('tag')
+        .withGraphJoined('tag')
+        .whereIn('key', keys)
```
TimCsaky, line 117: we need to ensure we have no duplicate keys, so a diff of tag records is just checking the key column

app/src/services/tag.js
```diff
@@ -7,42 +7,196 @@ const { Tag, VersionTag } = require('../db/models');
...
+        }
+        // else add to array for inserting
+        else {
+          newTags.push({ key: key.toLowerCase(), value: value });
+        }
+      });
+
+      // insert new tags
+      if (newTags.length) {
+        const newTagArr = await Tag.query(trx).insert(newTags).returning('*');
```
jujaga, line 194: Multiline this query. Also worthwhile naming this variable as newTagset or something - always want to describe what the array is about instead of calling the variable an array itself.

app/src/controllers/object.js
```diff
@@ -415,11 +416,10 @@ const controller = {
       } else {
         await storageService.deleteObjectTagging(data);
       }
-
       // update tags for version in DB
       await utils.trxWrapper(async (trx) => {
         const version = await versionService.get(data.versionId, objId, trx);
-        await tagService.updateTags(version.id, toLowerKeys(data.tags), userId, trx);
+        await tagService.dissociateTags(version.id, keysToRemove, userId, trx);
```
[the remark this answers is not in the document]
TimCsaky: updated code. dissociateTags() should accepts an array of lowercase key names (eg: ['ministry', 'priority'])

app/src/services/tag.js
```diff
@@ -7,42 +7,196 @@ const { Tag, VersionTag } = require('../db/models');
...
+      trx = etrx ? etrx : await Tag.startTransaction();
+      let response = [];
+
+      // delete joining records
+      response = await VersionTag.query(trx)
+        .allowGraph('tag')
+        .withGraphJoined('tag')
+        .whereIn('key', keys)
+        .modify('filterVersionId', versionId)
+        .delete();
```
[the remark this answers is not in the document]
TimCsaky: this works

jujaga (approved):

app/src/components/utils.js
```diff
@@ -21,6 +21,22 @@ const utils = {
...
+    const keys = Object.keys(obj);
+    let n = keys.length;
+    let newObj = {};
+    while (n--) {
+      let key = keys[n];
+      newObj[key.toLowerCase()] = obj[key];
+    }
+    return newObj;
+  },
+
```
jujaga, line 39: This can be written more idiomatically in declarative format:
```js
  /**
   * @function toLowerCaseKeys
   * @param {object} obj an arbitrary object
   * @returns {object} the object obj with all keys converted to lowercase
   */
  toLowerCaseKeys(obj) {
    return Object.fromEntries(Object.entries(obj).map(entry => {
      return [entry[0].toLowerCase(), entry[1]];
    })));
  }
```

app/src/components/utils.js
```diff
@@ -184,6 +200,18 @@ const utils = {
     return Object.entries(input).map(([k, v]) => ({ key: k, value: v }));
   },
 
+  /**
+   * @function getTagsByKeyValue
+   * get tag objects in array that have given key and value
+   * @param {object[]} tags and array of tags (eg: [{ key: 'a', value: '1'}, { key: 'b', value: '1'}]
+   * @param {string} key the string to match in the tag's `key` property
+   * @param {string} value the string to match in the tag's `value` property
+   * @returns and array of matching tag objects
```
jujaga, line 209:
```js
* @returns {object[]} an array of matching tag objects
```

app/src/services/tag.js
```diff
@@ -98,24 +99,42 @@ const service = {
...
-
-      // delete joining records
-      response = await VersionTag.query(trx)
-        .allowGraph('tag')
-        .withGraphJoined('tag')
-        .whereIn('tag.key', keys)
-        .modify('filterVersionId', versionId)
-        .delete();
+      let response = 0;
+      let count = 0;
```
jujaga, line 112: Count should be instantiated within the async tag foreach callback function instead of being scoped one level up. It isn't being used anywhere else outside of that.

kamorel (approved):

PR #73, opened 2022-09-10
Separate Metadata and Metadada joins in service

TimCsaky:
# Description

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2783

Similar to the PR 67. Try to separate metadata service calls.
with metadata edits a new version is created UNLESS versioning is disabled.
so the collision detection is a bit different.

I've put the process of creating the Metadata records into its own separate function that does collision detection..

The other function associateMetadata() can be used for both add, replace and delete.
I could have put in a separate function that does just add but it seems unnecessary.

We should do a lot of testing against this PR deployment.

PruneOrphanMetdata() will need to be called when we call the deleteObject() controller.. which can be done in a separate PR.

## Types of changes

## Checklist

### DIFF
--- a/app/src/components/utils.js
+++ b/app/src/components/utils.js
@@ -22,17 +22,6 @@ const utils = {
   },
 
   /**
-   * @function toLowerCaseKeys
-   * @param {object} obj an object with keys of mixed-case
-   * @returns {object} the input object with keys converted to lowercase
-   */
-  toLowerCaseKeys(obj){
-    return Object.fromEntries(Object.entries(obj).map(entry => {
-      return [entry[0].toLowerCase(), entry[1]];
-    }));
-  },
-
-  /**
    * @function delimit
    * Yields a string `s` that will always have a trailing delimiter. Returns an empty string if falsy.
    * @param {string} s The input string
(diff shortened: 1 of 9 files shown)


app/src/components/utils.js
```diff
@@ -196,15 +185,15 @@ const utils = {
   },
 
   /**
-   * @function getTagsByKeyValue
```
TimCsaky, line 199: this was a duplicate of function toLowerKeys() i added previously by mistake

app/src/components/utils.js
```diff
@@ -196,15 +185,15 @@ const utils = {
...
-   * get tag objects in array that have given key and value
-   * @param {object[]} tags and array of tags (eg: [{ key: 'a', value: '1'}, { key: 'b', value: '1'}]
-   * @param {string} key the string to match in the tag's `key` property
-   * @param {string} value the string to match in the tag's `value` property
-   * @returns {object[]} an array of matching tag objects
+   * @function getObjectsByKeyValue
```
TimCsaky, line 188: relatively simple array find function that we can use in a few places

app/src/db/models/tables/metadata.js
```diff
@@ -20,7 +21,15 @@ class Metadata extends Model {
           },
           to: 'version.id'
         }
-      }
+      },
+      versionMetadata: {
```
TimCsaky, line 25: not currently using this relation

kamorel (approved):

PR #76, opened 2022-09-21
Add endpoints for searching only metadata and tags

kamorel:
# Description

Adds two GET endpoints for querying the tag and endpoint tables. Both endpoints accept multiple parameters to allow narrowing down the search.

## Types of changes

## Checklist


app/src/services/object.js
```diff
@@ -76,6 +76,21 @@ const service = {
...
 
+  /**
+   * @function searchMetadata
+    * Search and filter for specific metadata keys
+   * @param {object} [params.metadata] Optional object of metadata keys to filter on
+   * @returns {Promise<object[]>} The result of running the find operation
+   */
+  searchMetadata: (params) => {
+    return Metadata.query()
+      .modify('filterKey', { metadata: params.metadata })
```
TimCsaky, line 87: minor point, it might simplify things if you just pass the params.metadata array, instead of nesting it in another object.

app/src/routes/v1/object.js
```diff
@@ -19,6 +19,16 @@ routes.get('/', requireDb, objectValidator.searchObjects, (req, res, next) => {
   objectController.searchObjects(req, res, next);
 });
 
+/** Search for tags */
```
TimCsaky, line 22: change `tags` to `metadata`

TimCsaky (dismissed):
minor changes. we should discuss again if we want to try and match on the value as well.. it adds complication. I'm starting to think users will only ever really need to provide a single param (key and optional value)

app/src/services/object.js
kamorel: Will leave as is for now. It follows the same patterns as searchObjects

app/src/services/object.js
```diff
@@ -76,6 +76,21 @@ const service = {
...
+   * @function searchMetadata
+    * Search and filter for specific metadata keys
+   * @param {object} [params.metadata] Optional object of metadata keys to filter on
+   * @returns {Promise<object[]>} The result of running the find operation
+   */
+  searchMetadata: (params) => {
+    return Metadata.query()
+      .modify('filterKeyValue', { metadata: params.metadata })
+      .then(result => result.map(row => {
+        // eslint-disable-next-line no-unused-vars
```
TimCsaky, line 89: just checking if you need this linting exception thing

kamorel: Removed it.

app/src/db/models/tables/tag.js
```diff
@@ -34,6 +34,47 @@ class Tag extends Model {
...
+        const subqueries = [];
+
+        let filterValues = false;
+
+        if (value.tag && Object.keys(value.tag).length) {
+          Object.entries(value.tag).forEach(([key, val]) => {
+            let q;
+
+            if (val.length) {
+              q = Tag.query().distinct('key', 'value').where('key', 'ilike', `%${key}%`).where('value', 'ilike', `%${val}%`);
```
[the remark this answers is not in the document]
kamorel: Updated query to this.

TimCsaky (approved):

PR #83, opened 2022-11-22
Implement Bucket Permission Enforcement

jujaga:
# Description

This PR adds and extends COMS to handle and enforce permissions on buckets and their cascade behavior to objects. This is needed in order to ensure users do not access things they shouldn't be able to access.

- dd158f7 Add /bucket permission and validation endpoint enforcement
- 7374b29 Fix broken controller and validator unit tests
- b42fca2 Refactor hasPermission to leverage utility permission checking function
- ecaa673 Implement advanced bucket permission creation/grant flow
- da9f094 Bifurcate controller permissions layer to handle buckets and objects
- 1eedcf3 Bifurcate service permissions layer to handle buckets and objects
- 967509c Disable keycloak resource role mapping support
- 6f5015a Split permissions into new subfolder hierarchy
- 37a1e76 Rename routes to router for codewide consistency
- a846973 Refactor validation layer to split permissions

[SHOWCASE-2822?](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2822)

## Types of changes

Bug fix (non-breaking change which fixes an issue)
New feature (non-breaking change which adds functionality)

Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] I have read the [CONTRIBUTING](CONTRIBUTING.md) doc
- [x] I have checked that unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #85, opened 2022-11-23
add bucket ids to create and read

loneil:
Signed-off-by: Lucas ONeil <lucasoneil@gmail.com>

# Description
Handle the bucket ID in object create and search operations.

For create, get the bucket ID out of the multi part form data to pass to the service.
For search, there was a typo in the param name, and added validation and param parsing.

Tested leaving the bucket ID out in the call for each of these as well to make sure it handles when it's not provided.

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -202,6 +202,14 @@ const controller = {
       const objects = [];
       const userId = await userService.getCurrentUserId(getCurrentIdentity(req.currentUser, SYSTEM_USER));
 
+      let bucketId = undefined;
+      bb.on('field', (fieldname, val) => {
+        // We only handle 'bucketId' at this point
+        if(fieldname === 'bucketId') {
+          bucketId = val;
+        }
+      });
+
       bb.on('file', (name, stream, info) => {
         const objId = uuidv4();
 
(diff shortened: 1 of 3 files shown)


app/src/controllers/object.js
```diff
@@ -202,6 +202,14 @@ const controller = {
       const objects = [];
       const userId = await userService.getCurrentUserId(getCurrentIdentity(req.currentUser, SYSTEM_USER));
 
+      let bucketId = undefined;
+      bb.on('field', (fieldname, val) => {
```
TimCsaky, line 206: just wondering if the bucketId is going to be a query parameter (eg: POST `/object?bucketId=oiuo-r3434r34-3d4rd4-3445rtf-3ewedwe`) can't you get it using `req.query.bucketId` instead of in the busboy field event ? i think we're only uploading to a single bucket.

loneil: That could be an option yeah. Worth discussing with @jujaga as he's moved this into his other PR too

jujaga: Slipstreamed into #83

jujaga:
Slipstreamed into #83

PR #95, opened 2022-12-05
Expose GET /object and /bucket endpoints

TimCsaky:
# Description

Add userId as an optional query parameters for both `GET /object` and `GET /bucket` endpoints.. so users can acquire "data" about the object even if they do not have explicit permissions for it.

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2948
see 2nd and 3rd changes in: https://gist.github.com/jujaga/ca344f79c81e05796af04b44b30038ca#problem-2-proposal

## Types of changes

## Checklist

### DIFF
--- a/app/src/components/utils.js
+++ b/app/src/components/utils.js
@@ -152,12 +152,12 @@ const utils = {
     return objectArray.reduce((acc, obj) => {
       // value of the 'property' attribute of obj
       const val = obj[property];
-      // does accumulator array contain an object with permissions matching obj
+      // does accumulator array have element with nested array containing current obj
       const el = acc.find((ob) => {
         return ob[group].some((p) => p[property] === val);
       });
       if (el) {
-        // add object to current object's permissions array
+        // add to current element's nested array
         el[group].push(obj);
       } else {
         // add to a new top level element in accumulator array
--- a/app/src/controllers/bucket.js
+++ b/app/src/controllers/bucket.js
@@ -2,11 +2,9 @@ const Problem = require('api-problem');
 const { UniqueViolationError } = require('objection');
 const { NIL: SYSTEM_USER } = require('uuid');
 
-const { AuthMode } = require('../components/constants');
 const errorToProblem = require('../components/errorToProblem');
 const {
   addDashesToUuid,
-  getAppAuthMode,
   isTruthy,
   mixedQueryToArray,
   getCurrentIdentity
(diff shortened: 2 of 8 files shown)


app/src/components/utils.js
```diff
@@ -152,12 +152,12 @@ const utils = {
     return objectArray.reduce((acc, obj) => {
       // value of the 'property' attribute of obj
       const val = obj[property];
-      // does accumulator array contain an object with permissions matching obj
+      // does accumulator array have element with nested array containing  current obj
```
jujaga, line 155: Style: Extra space between containing and current.

app/src/controllers/bucket.js
```diff
@@ -16,9 +14,7 @@ const { redactSecrets } = require('../db/models/utils');
 const { bucketService, storageService, userService } = require('../services');
 
 const SERVICE = 'ObjectService';
-
-const authMode = getAppAuthMode();
-const secretFields = ['secretAccessKey'];
+const secretFields = ['secretAccessKey', 'accessKeyId'];
```
jujaga, line 17: Style: Please alphabetize this array.

app/src/controllers/object.js
```diff
@@ -691,6 +687,7 @@ const controller = {
       const metadata = getMetadata(req.headers);
       const tagging = req.query.tagset;
       const params = {
+        userId: req.query.userId,
```
jujaga, line 690: This userId attribute should be declared after the bucketId attribute.

Also, if we're going to be adding userId as a queryable parameter, we're going to need to make sure that we are handling not only the singular case, but also the plural case at some point. This should have mixedQueryToArray, but IIRC we don't actually support that atm in the filterUserId modifier on the service/db layers though, so this may need to be dealt with as a future PR.

app/src/validators/bucket.js
```diff
@@ -36,6 +36,7 @@ const schema = {
 
   searchBuckets: {
     query: Joi.object({
+      userId: scheme.guid,
```
jujaga, line 39: For now, if this is a singular GUID, IIRC we should be using the uuidv4 scheme instead of scheme.guid, which permits one, or an array of guids.

app/src/validators/object.js
```diff
@@ -120,6 +120,7 @@ const schema = {
   searchObjects: {
     headers: type.metadata(0),
     query: Joi.object({
+      userId: scheme.guid,
```
jujaga, line 123: For now, if this is a singular GUID, IIRC we should be using the uuidv4 scheme instead of scheme.guid, which permits one, or an array of guids.

jujaga (commented):
Looks decent, but we need to figure out what we're intending on doing with the plurality of the userId query parameter and remain consistent with implementation.

dc-bcgov (approved):

jujaga (approved):

PR #98, opened 2022-12-08
Feat/2927 bucket object permissions

dc-bcgov:
# Description

Altering 'bucketPermissions.searchPermissions' endpoint to accept a boolean query parameter called 'objectPerms=true'. When this is provided, implicit buckets will also be listed, even though they have no permissions bound to them. API response is now an array of arrays instead of just an array.

## Checklist

### DIFF
--- a/app/src/services/bucketPermission.js
+++ b/app/src/services/bucketPermission.js
@@ -117,9 +117,9 @@ const service = {
   getBucketIdsWithObject: async (userId) => {
     return ObjectPermission.query()
       .select('bucketId')
-      .modify('filterUserId', userId)
+      .distinct('userId')
       .joinRelated('object')
-      .whereIn('userId', userId)
+      .modify('filterUserId', userId)
       .whereNotNull('bucketId')
       .then(response => response.map(entry => entry.bucketId));
   },
--- a/app/src/services/bucketPermission.js
+++ b/app/src/services/bucketPermission.js
@@ -114,12 +114,12 @@ const service = {
    * @param {string|string[]} [params.userId] Optional string or array of uuids representing the user
    * @returns {Promise<object>} The result of running the find operation
    */
-  getBucketIdsWithObject: async (userIds) => {
+  getBucketIdsWithObject: async (userId) => {
     return ObjectPermission.query()
       .select('bucketId')
-      .distinct('bucketId')
+      .modify('filterUserId', userId)
       .joinRelated('object')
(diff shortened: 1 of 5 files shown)


app/src/services/bucketPermission.js
```diff
@@ -108,6 +108,22 @@ const service = {
...
+   * Searches for specific (bucket) object permissions
+   * @param {string|string[]} [params.userId] Optional string or array of uuids representing the user
+   * @returns {Promise<object>} The result of running the find operation
+   */
+  getBucketIdsWithObject: async (userIds) => {
+    return ObjectPermission.query()
+      .select('bucketId')
+      .distinct('bucketId')
+      .joinRelated('object')
+      .whereIn('userId', userIds)
```
jujaga, line 122: Let's replace this line with `.modify('filterUserId', userIds)`. The way this is written right now, if you pass in a string to `userId`, it won't work. It only looks like it works because we already pre-cast it in the controller this is invoked in with the `mixedQueryToArray` function, which always makes everything an array, even if there's only one element. We should be able to support both a string, or an array of strings just like what the JSDocs specify on line 115.

app/src/services/bucketPermission.js
```diff
@@ -108,6 +108,22 @@ const service = {
     }
   },
 
+  /**
+   * @function getBucketIdsWithObject
+   * Searches for specific (bucket) object permissions
+   * @param {string|string[]} [params.userId] Optional string or array of uuids representing the user
+   * @returns {Promise<object>} The result of running the find operation
+   */
+  getBucketIdsWithObject: async (userIds) => {
```
jujaga, line 117: `userIds` should be singular `userId`. While it can accept one or many, we generally stick with singular naming unless the parameter must always contain multiple values.

app/src/validators/bucketPermission.js
```diff
@@ -7,7 +7,8 @@ const schema = {
     query: Joi.object({
       userId: scheme.guid,
       bucketId: scheme.guid,
-      permCode: scheme.permCode
+      permCode: scheme.permCode,
+      objectPerms: type.truthy
```
jujaga, line 11: Style: Alphabetize these attributes; o < p.

TimCsaky (commented):
Tested the new code locally and works as expected.

TimCsaky (approved):

jujaga (approved):

PR #104, opened 2022-12-21
Updating (swagger) api spec to comply with current endpoint(s) implementation

dc-bcgov:
# Description
The api spec 'v1.api-spec.yaml' file has been updated to comply with the current endpoints implementation. 

## Types of changes

> New tags were added: 'Bucket', 'Bucket Permission' and 'Object Permission'.
> New paths were added: '/bucket', '/bucket/{bucketId}', '/permission/object', '/permission/object/{objectId}', '/permission/bucket', and '/permission/bucket/{bucketId}'.
> The all other paths were verified and confirmed they are up to date.

## Checklist

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -164,8 +164,7 @@ paths:
 
     patch:
       summary: Updates a bucket
-      description: >-
-        Returns updated bucket
+      description: Returns updated bucket
       operationId: updateBucket
       tags:
         - Bucket
@@ -189,8 +188,7 @@ paths:
 
     delete:
       summary: Deletes a bucket
-      description: >-
-        Delete bucket based on bucketId
+      description: Delete bucket based on bucketId
       operationId: deleteBucket
       tags:
         - Bucket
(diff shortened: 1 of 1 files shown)


app/src/docs/v1.api-spec.yaml
```diff
@@ -1049,6 +1375,47 @@ components:
...
+            endpoint:
+              type: string
+            key:
+              type: string
+              description: The prefix to path of all objects in the bucket
+              example: coms/env
+            region:
+              type: string
+            secretAccessKey:
+              example: REDACTED
```
jujaga, line 1417: When defining properties, make sure to always include a description and an example wherever possible. While this is "sufficient" for OpenAPI spec definition, we generally like to go the extra mile and make sure everything is relentlessly documented especially for component definitions.

app/src/docs/v1.api-spec.yaml
```diff
@@ -1135,6 +1502,54 @@ components:
...
+              type: string
+              format: uuid
+              description: The unique identifier of the user
+              example: ac246e31-c807-496c-bc93-cd8bc2f1b2b4
+            updatedAt:
+              type: string
+              format: yyyy-mm-ddThh:mm:ssZ
+              description: The created by date
+              example: 2022-12-02T18:13:09.954Z
+        - $ref: '#/components/schemas/DB-TimestampUserData'
```
jujaga, line 1552: These 4 attributes do not need to be re-declared as you're already pulling in `#/components/schemas/DB-TimestampUserData` on Line 1552.

PR #106, opened 2022-12-30
Add deleteMarker and latest object search params

TimCsaky:
# Description

allow for object search filtering by 
- soft-deleted objects (version.deleteMarker)
- latest version (when used in conjunction with filters on metadata, tags and deleteMarker)

implementation details:
- separate query params
- separate modifiers
- updated api spec

sql generated with `?latest=true&deleteMarker=true` :

```sql
SELECT 
 *
FROM
	"object"
LEFT JOIN (
	SELECT
		DISTINCT ON
		("version"."objectId") "version" .*
	FROM
		"version"
	ORDER BY
		"version"."objectId" ASC,
		"version"."createdAt" DESC) AS "version" ON
	"version"."objectId" = "object"."id"
WHERE
	"version"."deleteMarker" = true
```

## Checklist

### DIFF
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -701,7 +701,9 @@ const controller = {
         metadata: metadata && Object.keys(metadata).length ? metadata : undefined,
         tag: tagging && Object.keys(tagging).length ? tagging : undefined,
         public: isTruthy(req.query.public),
-        active: isTruthy(req.query.active)
+        active: isTruthy(req.query.active),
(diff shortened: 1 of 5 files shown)


TimCsaky:
I kind of wonder if one day we might get the time to consider how these search modifiers are layer.
it might make sense to have a 'base' select query that joins all the child tables.
And then have each modifier do more of a filtering process rather than having `WHERE version.is IN (<result of subquery>)..`

jujaga (approved):

dc-bcgov (approved):

PR #108, opened 2023-01-04
Extending get object permissions api

dc-bcgov:
Extending get object permissions endpoint by adding 'bucketPerms' flag to include all object permissions within bucket if flag set to true.

# Description

## Checklist

### DIFF
--- a/app/src/controllers/objectPermission.js
+++ b/app/src/controllers/objectPermission.js
@@ -31,6 +31,20 @@ const controller = {
         permCode: utils.mixedQueryToArray(req.query.permCode)
       });
       const response = utils.groupByObject('objectId', 'permissions', result);
+
+      if (utils.isTruthy(req.query.bucketPerms)) {
+        const objectIds = await objectPermissionService.getObjectIdsWithBucket(userId);
+
+        objectIds.forEach(objectId => {
+          if (!response.map(r => r.objectId).includes(objectId)) {
+            response.push({
+              objectId: objectId,
+              permissions: []
+            });
+          }
+        });
+      }
+
       res.status(200).json(response);
     } catch (e) {
       next(errorToProblem(SERVICE, e));
--- a/app/src/controllers/objectPermission.js
+++ b/app/src/controllers/objectPermission.js
@@ -33,7 +33,7 @@ const controller = {
       const response = utils.groupByObject('objectId', 'permissions', result);
 
       if (utils.isTruthy(req.query.bucketPerms)) {
-        const objectIds = await objectPermissionService.getObjectIdsWithBucket(userId);
+        const objectIds = await objectPermissionService.getObjectIdsWithBucket(userId, bucketIds);
 
         objectIds.forEach(objectId => {
           if (!response.map(r => r.objectId).includes(objectId)) {
--- a/app/src/services/objectPermission.js
+++ b/app/src/services/objectPermission.js
@@ -1,7 +1,7 @@
 const { v4: uuidv4, NIL: SYSTEM_USER } = require('uuid');
 
 const { Permissions } = require('../components/constants');
-const { ObjectPermission } = require('../db/models');
+const { ObjectPermission, BucketPermission } = require('../db/models');
 
 /**
  * The Object Permission DB Service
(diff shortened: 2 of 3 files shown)


app/src/services/objectPermission.js
```diff
@@ -61,6 +61,24 @@ const service = {
...
 
+  /**
+   * @function getObjectIdsWithBucket
+   * Searches for specific (object) bucket permissions
+   * @param {string|string[]} [params.userId] Optional string or array of uuids representing the user
+   * @param {string|string[]} [params.bucketId] Optional string or array of bucket id(s)
+   * @returns {Promise<object>} The result of running the find operation
+  */
+  getObjectIdsWithBucket: async (userId, bucketId) => {
+    return ObjectPermission.query()
```
TimCsaky, line 72: here we want to get 'all objects that are in buckets that user has any bucket permission for'. 
i believe we can do:
```js
    return BucketPermission.query()
   .rightJoin('object', 'bucket_permission.bucketId', '=', 'object.bucketId')
  .select('object.objectId')
  .distinct('userId')
  .modify('filterUserId', userId)
  .modify('filterBucketId', bucketId)
  .whereNotNull('object.bucketId')
  .then(response => response.map(entry => entry.objectId));

File: app/src/controllers/objectPermission.js [hunk 2]
```diff
@@ -31,6 +31,20 @@ const controller = {
         permCode: utils.mixedQueryToArray(req.query.permCode)
       });
       const response = utils.groupByObject('objectId', 'permissions', result);
+
+      if (utils.isTruthy(req.query.bucketPerms)) {
+        const objectIds = await objectPermissionService.getObjectIdsWithBucket(userId, bucketIds);
```
TimCsaky, line 36: note: the `bucketIds` param is used to optionally scope the object permission search to one(or more) buckets, if undefined (not passed as an input param) the search is for object perms across all buckets.

To include objects with implicit (bucket-level permission) in the response, by passing in `bucketPerms=true`, i think it's good to pass the `bucketIds` param on to the `getObjectIdsWithBucket()` function.

TimCsaky (approved):

app/src/services/objectPermission.js
```diff
@@ -1,7 +1,7 @@
 const { v4: uuidv4, NIL: SYSTEM_USER } = require('uuid');
 
 const { Permissions } = require('../components/constants');
-const { ObjectPermission } = require('../db/models');
+const { ObjectPermission, BucketPermission } = require('../db/models');
```
jujaga, line 4: Style: Alphabetical ordering of object imports

app/src/services/objectPermission.js
```diff
@@ -61,6 +61,22 @@ const service = {
...
+   * @param {string|string[]} [params.userId] Optional string or array of uuids representing the user
+   * @param {string|string[]} [params.bucketId] Optional string or array of bucket id(s)
+   * @returns {Promise<object>} The result of running the find operation
+  */
+  getObjectIdsWithBucket: async (userId, bucketId) => {
+    return BucketPermission.query()
+      .distinct('object.id AS objectId')
+      .rightJoin('object', 'bucket_permission.bucketId', '=', 'object.bucketId')
+      .modify('filterUserId', userId)
+      .whereIn('bucket_permission.bucketId', bucketId)
```
jujaga, line 76: Double check but I think this could be a modify `filterBucketId` call.

TimCsaky: there is but it caused ambiguous ref to bucketId.

jujaga: Change the ordering of the statements so that they're logically grouped then:
```js
      .distinct('object.id AS objectId')
      .modify('filterUserId', userId)
      .rightJoin('object', 'bucket_permission.bucketId', '=', 'object.bucketId')
      .whereIn('bucket_permission.bucketId', bucketId)
```

jujaga: Also may be worth using `.rightJoinRelated` instead of rightJoin as it'll let us leverage the ORM relation mappings instead.
https://vincit.github.io/objection.js/api/query-builder/join-methods.html#rightjoinrelated

jujaga (approved):

PR #110, opened 2023-01-17
Joi Validation updates

TimCsaky:
# Description

- changes to the order of middleware on routes so validation is done before auth
- Updates to Joi validation rules and a few missing Joi schema unit tests
- api spec update coming in next sprint

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -707,7 +707,7 @@ const controller = {
       };
 
       const response = await objectService.searchObjects(params);
-      res.status(201).json(response);
+      res.status(200).json(response);
     } catch (error) {
       next(error);
     }
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -597,37 +597,30 @@ const controller = {
       }
 
       const newMetadata = getMetadata(req.headers);
-      if (!Object.keys(newMetadata).length) {
-        // TODO: Validation level logic. To be moved.
-        // 422 when no keys present
-        res.status(422).end();
-      }
-      else {
-        const data = {
-          copySource: objPath,
-          filePath: objPath,
-          metadata: {
-            name: source.Metadata.name,  // Always enforce name and id key behavior
-            ...newMetadata, // Add new metadata
-            id: source.Metadata.id
-          },
-          metadataDirective: MetadataDirective.REPLACE,
-          versionId: sourceVersionId
-        };
-        const s3Response = await storageService.copyObject(data);
+      const data = {
+        copySource: objPath,
+        filePath: objPath,
+        metadata: {
+          name: source.Metadata.name,  // Always enforce name and id key behavior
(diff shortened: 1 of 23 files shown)


app/src/validators/common.js
```diff
@@ -45,11 +45,12 @@ const type = {
...
 
-  metadata: (min) => Joi.object()
-    .pattern(/^x-amz-meta-.{1,255}$/i, Joi.string().min(min).max(255))
+  metadata: (min = 0) => Joi.object()
+    .pattern(/^x-amz-meta-.{1,255}$/i, Joi.string().min(1).max(255), { matches: Joi.array().min(min) })
     .unknown(),
 
-  tagset: (min) => Joi.object().pattern(/^.{1,128}$/, Joi.string().min(min).max(255))
+  tagset: (min = 0) => Joi.object()
+    .pattern(/^.{1,128}$/, Joi.string().min(1).max(255), { matches: Joi.array().min(min) })
```
pbolduc, line 53: Tag values can be up to [256 characters](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html)

TimCsaky: Hey Phil. Glad to see you're still following the repo.
Good to know about the character limit. I think we limited this to 255 because the field in the db where we store this value is limited to 255. Next time we do a db migration we'll consider increasing. Agree, it does seem a little odd in comparison to the S3 docs.

app/src/middleware/authorization.js
```diff
@@ -100,6 +100,17 @@ const currentObject = async (req, _res, next) => {
   next();
 };
 
+/**
+ * @function hasPermission
+ * fucntion checks:
```
TimCsaky, line 105: typo

app/src/controllers/object.js
```diff
@@ -845,7 +838,7 @@ const controller = {
           ...dbResponse,
           ...s3Response
         };
-        res.status(200).json(result);
+        res.status(201).json(result);
```
TimCsaky, line 841: keep as 200

app/src/validators/common.js
pbolduc: It is also documented as 256 in your [Metadata-Tag](https://github.com/bcgov/common-object-management-service/wiki/Metadata-Tag#tag) wiki page.

jujaga (approved):

app/src/validators/common.js
jujaga: We've added this discrepancy to our internal backlog and will address it before the v0.4.0 release.

pbolduc: I would also comment, that metadata keys must conform to the [valid http header spec](https://www.rfc-editor.org/rfc/rfc2616#section-4.2). Cant contain control characters, or any of: ()<>@,;:\"/[]?={} _space_ and _tab_

loneil (approved):

PR #135, opened 2023-03-02
Improve unit test coverage to above 50%

dc-bcgov:
This is a first set of unit tests for service layer. New happy path unit tests have been added for object, objectPermission, bucket and bucketPermission services.

# Description

## Checklist


app/tests/unit/services/object.spec.js
```diff
@@ -0,0 +1,100 @@
...
+  MockModel.mockReset();
+  MockTransaction.mockReset();
+});
+
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
+describe('bucket', () => {
+  const createSpy = jest.spyOn(service, 'create');
```
TimCsaky, line 42: Because we want to test the service calls. we dont actually want to mock these.
other wise they will never run. see comment below

app/tests/unit/services/object.spec.js
```diff
@@ -0,0 +1,100 @@
...
+    searchObjectsSpy.mockRestore();
+    getBucketKeySpy.mockRestore();
+    readSpy.mockRestore();
+    updateSpy.mockRestore();
+  });
+
+  it('Create object', async () => {
+    const etrx = await jest.fn().mockResolvedValue(MockTransaction);
+    await service.create(data, etrx);
+    expect(createSpy).toHaveBeenCalledTimes(1);
```
TimCsaky, line 70: here, we dont want to check the createSpy (service call) has been called.
instead, we should test the logic, whats happening inside the create() service function.
for example:
- mock the the `ObjectModel.query..`  function and check that has been called
- set data.userId to undefined and check that `objectPermissionService.addPermissions()` has been called with undefined.

TimCsaky: that Coverage Gutters vs code plugin should show you which lines of code need testing.

jujaga (commented):
While this did improve code coverage, we still want to make sure that we're testing to make sure certain execution outputs are coming out with expected results. Make sure to retrace through the services being called and make sure that the general logic and expected outputs are being asserted as a baseline.

app/tests/common/dbHelper.js
```diff
@@ -33,22 +33,25 @@ MockTransaction.mockReset = () => {
...
 MockModel.findById = jest.fn().mockReturnThis();
 MockModel.first = jest.fn().mockReturnThis();
 MockModel.delete = jest.fn().mockReturnThis();
 MockModel.deleteById = jest.fn().mockReturnThis();
+MockModel.distinct = jest.fn().mockReturnThis();
 MockModel.insert = jest.fn().mockReturnThis();
 MockModel.insertAndFetch = jest.fn().mockReturnThis();
 MockModel.insertGraph = jest.fn().mockReturnThis();
 MockModel.insertGraphAndFetch = jest.fn().mockReturnThis();
-MockModel.findById = jest.fn().mockReturnThis();
```
TimCsaky, line 44: I think we still need findById() mocked because it is used elsewhere.

app/tests/unit/services/bucketPermission.spec.js
```diff
@@ -0,0 +1,80 @@
...
+  });
+});
+
+describe('addPermissions', () => {
+
+  it('should search and return bucket permissions', async () => {
+    // add userId property
+    const dataWithUser = { ...data, userId: userId };
+
+    const perms = Object.values(Permissions).map((p) => ({
```
TimCsaky, line 49: i dont think we need this perms object. searchPermissionSpy should return bucket permissions in a different format.
eventually we will want to test the code between line 34 and 48 of the addPermission() function.
maybe just mock everything in this addPermissions function and dont bother testing anything except that mockeModel was called.

app/tests/unit/services/bucketPermission.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    const etrx = await jest.fn().mockResolvedValue(MockTransaction);
+    await service.addPermissions(objectId, data, {currentUserId: SYSTEM_USER}, etrx);
+    expect(MockModel.query).toHaveBeenCalledTimes(1);
+  });
+
+});
+
+describe('removePermissions', () => {
+
+  it('should delete bucket permissions', async () => {
```
TimCsaky, line 66: if these are failing, skip them

app/tests/unit/services/bucketPermission.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    await service.removePermissions(objectId);
+    expect(MockModel.query).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe('getBucketIdsWithObject', () => {
+
+  it('should return bucket ids with object', async () => {
```
TimCsaky, line 74: if these are failing, skip them

app/tests/unit/services/object.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    expect(addPermissionSpy).toHaveBeenCalledTimes(1);
+    // addPermissions was called with expected paramters
+    expect(addPermissionSpy).toHaveBeenCalledWith(objectId, perms, userId, etrx);
+    // expect(MockModel.query).toHaveBeenCalledTimes(1);
+  });
+
+  it.skip('should add object permissions, no transaction provided', async () => {
+
+    // add userId property
+    const dataWithUser = { ...data, userId: userId };
```
TimCsaky, line 75: we dont need to add the userId, that was for the previous test. can delete this line.
and call `await service.create(data);` instead below

app/tests/unit/services/bucketPermission.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    service.searchPermissions(params);
+    expect(MockModel.query).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe('addPermissions', () => {
+
+  it('should search and return bucket permissions', async () => {
+    // add userId property
+    const dataWithUser = { ...data, userId: userId };
```
TimCsaky, line 47: we shouldnt keep doing this. instead you can redefine the data object as needed

app/tests/unit/services/objectPermission.spec.js
```diff
@@ -0,0 +1,94 @@
...
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
+
+describe('addPermissions', () => {
+
+  it('should search and add object permissions', async () => {
+    // add userId property
+    const dataWithUser = { ...data, userId: userId };
```
TimCsaky, line 47: we dont want to do this. that was just to check the `if data.userId(...` logic in another function.
you can define the params you need within the scope of this test block.

app/tests/unit/services/objectPermission.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+    const perms = Object.values(Permissions).map((p) => ({
```
TimCsaky, line 49: as above. you dont need this

TimCsaky (commented):
as discussed, there is definitely a learning process with these Jest unit tests.
I would recommend removing pretty much everything except the mockModel calls.. and skip the tests that are failing.

jujaga:
Setting as draft - to be revisited at a later date.

PR #148, opened 2023-03-31
Feature/validate bucket creds

loneil:
# Description
Need feedback on implementation. If we think this is the right spot for this I'll add test coverage.

Think we should go with a more generic 409 here. Looked up 406 and that's about "content negotiation" with the request headers mismatching what would be returned, which doesn't really apply here.

All screenshots below from BCBox are just generic handlers for errors, specific UX/wording to come in later ticket (unless this suffices).

**Changes**
On the bucket create and update routes use the existing headBucket service method to throw an exception if the object store/bucket creds fail.
Alter that headBucket method to take in the desired bucket config rather than look it up (in create and update scenarios) so that it tries the proposed changes to the bucket rather than looking up config or DB (since that would obviously pass regardless of what the user is trying to cram in)

Exception will set 409 as response code and then...
image

(as said, can adjust this in BCBox ticket as we can handle a 409 in a specific way if needed)

**Error handling**
The sdk throws (totally) different errors depending on which way the user screws up their config. So we could handle each of this different formats and return different errors but I propose (and implemented) we just catch all and return a generic message (see code in `_validateCredentials`) because
- Obvious security pattern of not telling an attacker what part is wrong in their auth attempt
- Maybe the sdk changes error format in the future or something?

Invalid key/secret
image

Wrong object store URL
image

Wrong Bucket
image

## Checklist

### DIFF
--- a/app/src/components/constants.js
+++ b/app/src/components/constants.js
@@ -21,6 +21,9 @@ module.exports = Object.freeze({
     NONE: 'NONE'
   },
 
+  /** Need to specify valid AWS region or it'll explode ('us-east-1' is default, 'ca-central-1' for Canada) */
+  DEFAULTREGION: 'us-east-1',
+
   /** Download mode behavior overrides */
   DownloadMode: {
     /** Proxies payload data through COMS */
--- a/app/src/controllers/bucket.js
+++ b/app/src/controllers/bucket.js
@@ -15,7 +15,7 @@ const { redactSecrets } = require('../db/models/utils');
 
 const { bucketService, storageService, userService } = require('../services');
 
-const SERVICE = 'ObjectService';
+const SERVICE = 'BucketService';
 const secretFields = ['accessKeyId', 'secretAccessKey'];
 
 /**
@@ -67,7 +67,7 @@ const controller = {
 
   /**
    * @function _validateCredentials
-   * Gaurd against creating or update a bucket with invalid creds
+   * Guard against creating or update a bucket with invalid creds
    * @param {object} requestBody The body of the request
    */
   async _validateCredentials(requestBody, res) {
@@ -77,10 +77,10 @@ const controller = {
         bucket: requestBody.bucket,
         endpoint: requestBody.endpoint,
         key: requestBody.key,
-        region: DEFAULTREGION,
+        region: requestBody.region || DEFAULTREGION,
         secretAccessKey: requestBody.secretAccessKey,
       };
-      await storageService.headBucket(undefined, bucketSettings);
+      await storageService.headBucket(bucketSettings);
     } catch (e) {
       // If it's caught here it's unable to validate the supplied store/bucket and creds
       log.warn(`Failure to validate bucket credentials: ${e.message}`, {
@@ -106,7 +106,7 @@ const controller = {
     let response = undefined;
 
     // Check for credential accessibility/validity first
-    await this._validateCredentials(data, res);
+    await controller._validateCredentials(data, res);
 
     try {
       data.userId = await userService.getCurrentUserId(getCurrentIdentity(req.currentUser, SYSTEM_USER));
@@ -156,7 +156,7 @@ const controller = {
   async headBucket(req, res, next) {
     try {
       const bucketId = addDashesToUuid(req.params.bucketId);
-      await storageService.headBucket(bucketId);
+      await storageService.headBucket({ bucketId });
 
       res.status(204).end();
     } catch (e) {
@@ -217,7 +217,7 @@ const controller = {
    */
   async updateBucket(req, res, next) {
     // Check for credential accessibility/validity first
-    await this._validateCredentials(req.body, res);
+    await controller._validateCredentials(req.body, res);
 
     try {
       const userId = await userService.getCurrentUserId(getCurrentIdentity(req.currentUser, SYSTEM_USER), SYSTEM_USER);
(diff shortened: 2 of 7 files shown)


app/src/components/constants.js
```diff
@@ -21,6 +21,9 @@ module.exports = Object.freeze({
     NONE: 'NONE'
   },
 
+  /** Need to specify valid AWS region or it'll explode ('us-east-1' is default, 'ca-central-1' for Canada) */
+  DEFAULTREGION: 'us-east-1',
```
dc-bcgov, line 25: Wondering do we want to have this defined in config file instead? As in utils/getBucket the rest of the config values are being read from the config file.

app/src/controllers/bucket.js
```diff
@@ -63,6 +65,34 @@ const controller = {
...
+   * @param {object} requestBody The body of the request
+   */
+  async _validateCredentials(requestBody, res) {
+    try {
+      const bucketSettings = {
+        accessKeyId: requestBody.accessKeyId,
+        bucket: requestBody.bucket,
+        endpoint: requestBody.endpoint,
+        key: requestBody.key,
+        region: DEFAULTREGION,
```
jujaga, line 80: Region will need to be accepted as an optional requestBody parameter at some point; defaulting to DEFAULTREGION if undefined.
```js
region: requestBody.region || DEFAULTREGION,
```

app/src/components/constants.js
loneil: That might be a good idea for other people hosting their own COMS down the line yeah indeed.
For this just moving the constant for use in more than 1 spot but think that's worth making a backlog item for investigating at least.

app/src/controllers/bucket.js
```diff
@@ -74,9 +104,13 @@ const controller = {
   async createBucket(req, res, next) {
     const data = { ...req.body };
     let response = undefined;
+
+    // Check for credential accessibility/validity first
+    await this._validateCredentials(data, res);
```
dc-bcgov, line 109: Do we want this call outside of try/catch?

app/src/services/storage.js
```diff
@@ -193,10 +193,14 @@ const objectStorageService = {
...
+   * @param {Object} [bucketSettings] Optional bucketSettings for checking new additions/updates
    * @returns {Promise<object>} The response of the head bucket operation
    */
-  async headBucket(bucketId = undefined) {
-    const data = await utils.getBucket(bucketId, true);
+  async headBucket(bucketId = undefined, bucketSettings = null) {
+    // If already providing bucket settings don't go and get them from the DB or configuration
+    const data = bucketSettings
+      ? bucketSettings
+      : await utils.getBucket(bucketId, true);
```
jujaga, line 203: I think we should change this function signature to accept a generic options object instead of having two argument types, and then discern/prioritize logic based on the existence/lack of bucketId attribute:

```js
/**
 * @function headBucket
 * Checks if a bucket exists and if the S3Client has correct access permissions
 * You must supply either a `bucketId`, or the full S3 client credentials to test against
 * @param {string} [options.bucketId] Optional bucketId
 * @param {string} [options.accessKeyId] Optional S3 accessKeyId
 * @param {string} [options.bucket] Optional S3 bucket
 * @param {string} [options.endpoint] Optional S3 endpoint
 * @param {string} [options.key] Optional S3 key/prefix
 * @param {string} [options.region] Optional S3 region
 * @param {string} [options.secretAccessKey] Optional S3 secretAccessKey
 * @returns {Promise<HeadBucketCommandOutput>} The response of the head bucket operation
 */
async headBucket(options = {}) {
  const data = options.bucketId
    ? await utils.getBucket(options.bucketId, true)
    : options;
  const params = {
    Bucket: data.bucket,
  };

  return this._getS3Client(data).send(new HeadBucketCommand(params));
}
```

app/src/components/constants.js
jujaga: The full intent of COMS is that it will eventually (if not already doing so) accept a user-specified region. This should only be in config IF and ONLY IF the bucket must specify its precise region for the default bucket scenario; otherwise this should default to the AWS default of `us-east-1`, which is why it's treated more as an in-code constant.

app/src/controllers/bucket.js
jujaga: In this flow, it's leveraging the global catch resolver in app.js to handle the Problem object that has been thrown. The global catch resolver knows how to handle Problem object types and return a request appropriately, so a try/catch wrapper is not necessary here.

dc-bcgov (approved):

app/src/components/constants.js
loneil: Yeah, worth tracking, agree the call specifying it will be useful when that gets implemented (other comment in here).
If anyone is using the default bucket mode that would be a good place to take it into consideration for config.

app/src/controllers/bucket.js
loneil: Yeah exactly. The local catch could be used but then it would have to have logic to detect what the head bucket stuff throws (see notes in discussion about the different exception cases from that), or just rethrow the 409.
So think it's simpler and more maintainable to throw outside the one in the function, mostly because that local catch already does its own additional logic so don't like the catch block getting too big.

Doing the `await controller._validateCredentials(data, res);`

app/src/controllers/bucket.js
loneil: Done

jujaga (approved):

kamorel (approved):

PR #149, opened 2023-04-01
Feature/secret override

cberg-aot:
# Description

## Checklist

### DIFF
--- a/charts/coms/templates/secret.yaml
+++ b/charts/coms/templates/secret.yaml
@@ -48,7 +48,7 @@ data:
   password: {{ .Values.dbSecretOverride.password | default $pPassword | quote }}
   username: {{ .Values.dbSecretOverride.username | default $pUsername | quote }}
 {{- end }}
-{{- if not $oSecret }}
+{{- if and (not $oSecret) (and .Values.objectStorageSecretOverride.password .Values.objectStorageSecretOverride.username) }}
 ---
 apiVersion: v1
 kind: Secret
@@ -64,7 +64,7 @@ data:
   password: {{ .Values.objectStorageSecretOverride.password | default $oPassword | quote }}
   username: {{ .Values.objectStorageSecretOverride.username | default $oUsername | quote }}
 {{- end }}
-{{- if not $kSecret }}
+{{- if and (not $kSecret) (and .Values.keycloakSecretOverride.password .Values.keycloakSecretOverride.username) }}
 ---
 apiVersion: v1
 kind: Secret
--- a/charts/coms/templates/secret.yaml
+++ b/charts/coms/templates/secret.yaml
@@ -2,37 +2,61 @@
 {{- $bUsername := (randAlphaNum 32) | b64enc }}
 {{- $pPassword := (randAlphaNum 32) | b64enc }}
 {{- $pUsername := (randAlphaNum 32) | b64enc }}
+{{- $oPassword := (randAlphaNum 32) | b64enc }}
+{{- $oUsername := (randAlphaNum 32) | b64enc }}
 
 {{- $bSecretName := printf "%s-%s" (include "coms.fullname" .) "basicauth" }}
 {{- $bSecret := (lookup "v1" "Secret" .Release.Namespace $bSecretName ) }}
 {{- $pSecretName := printf "%s-%s" (include "coms.fullname" .) "passphrase" }}
 {{- $pSecret := (lookup "v1" "Secret" .Release.Namespace $pSecretName ) }}
+{{- $oSecretName := printf "%s-%s" (include "coms.fullname" .) "objectstorage" }}
+{{- $oSecret := (lookup "v1" "Secret" .Release.Namespace $oSecretName ) }}
 
 {{- if not $bSecret }}
 ---
 apiVersion: v1
 kind: Secret
 metadata:
+  {{- if not .Values.config.releaseScoped }}
   annotations:
     "helm.sh/resource-policy": keep
+  {{- end }}
   name: {{ $bSecretName }}
   labels: {{ include "coms.labels" . | nindent 4 }}
 type: kubernetes.io/basic-auth
 data:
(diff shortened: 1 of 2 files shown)


charts/coms/templates/secret.yaml
```diff
@@ -13,26 +21,62 @@
...
     "helm.sh/resource-policy": keep
+  {{- end }}
   name: {{ $bSecretName }}
   labels: {{ include "coms.labels" . | nindent 4 }}
 type: kubernetes.io/basic-auth
 data:
-  password: {{ $bPassword }}
-  username: {{ $bUsername }}
+  password: {{ .Values.basicAuthSecretOverride.password  | default $bPassword | quote }}
+  username: {{ .Values.basicAuthSecretOverride.username  | default $bUsername | quote }}
```
jujaga, line 33: Style: extra whitespace before the |

charts/coms/templates/secret.yaml
```diff
@@ -13,26 +21,62 @@
...
     "helm.sh/resource-policy": keep
+  {{- end }}
   name: {{ $pSecretName }}
   labels: {{ include "coms.labels" . | nindent 4 }}
 type: kubernetes.io/basic-auth
 data:
-  password: {{ $pPassword }}
-  username: {{ $pUsername }}
+  password: {{ .Values.dbSecretOverride.password  | default $pPassword | quote }}
+  username: {{ .Values.dbSecretOverride.username  | default $pUsername | quote }}
```
jujaga, line 49: Style: extra whitespace before the |

charts/coms/templates/secret.yaml
```diff
@@ -13,26 +21,62 @@
...
+  {{- if not .Values.config.releaseScoped }}
+  annotations:
+    "helm.sh/resource-policy": keep
+  {{- end }}
+  name: {{ $oSecretName }}
+  labels: {{ include "coms.labels" . | nindent 4 }}
+type: kubernetes.io/basic-auth
+data:
+  password: {{ .Values.objectStorageSecretOverride.password  | default $oPassword | quote }}
+  username: {{ .Values.objectStorageSecretOverride.username  | default $oUsername | quote }}
```
jujaga, line 65: Style: extra whitespace before the |

jujaga (dismissed):
On visual inspection this looks correct. Other than a minor whitespace style issue, this looks good otherwise. Please tweak that, and make sure to rebase branch to the latest upstream changes (PR-148 just got merged in this morning) and it should be good to go. 👍 
Pre-emptively signing off on this with the acknowledgement that the minor style changes will be done.

jujaga (changes requested):
Just thought through your PR a bit more. I think the original reason why we didn't include the `keycloak` and `objectstorage` secret definitions was that it does not really ever make sense to auto-generate them as they're externally defined secrets. Would you be able to modify your PR so that in addition to checking if the secret does not already exist, to only execute and apply those two secrets IF and ONLY IF you have the respective override values defined in your values? (Ref `charts/coms/templates/secret.yaml` Lines 51 and 67 for where I think we'll need more conditional checks)

cberg-aot:
I did a manual smell check of the conditional templating logic.  

I played with logic:
 - (or username password) 
 - (and username password) 
Autogenerating a default password/username could still be supported using or gate.

charts/coms/templates/secret.yaml
```diff
@@ -1,10 +1,18 @@
 {{- $bPassword := (randAlphaNum 32) | b64enc }}
 {{- $bUsername := (randAlphaNum 32) | b64enc }}
+{{- $kPassword := (randAlphaNum 32) | b64enc }}
+{{- $kUsername := (randAlphaNum 32) | b64enc }}
+{{- $oPassword := (randAlphaNum 32) | b64enc }}
+{{- $oUsername := (randAlphaNum 32) | b64enc }}
```
jujaga, line 6: We can probably drop these as I do not believe there are any circumstances where we would ever want to use this.

charts/coms/templates/secret.yaml
```diff
@@ -13,26 +21,62 @@
...
+  {{- if not .Values.config.releaseScoped }}
+  annotations:
+    "helm.sh/resource-policy": keep
+  {{- end }}
+  name: {{ $oSecretName }}
+  labels: {{ include "coms.labels" . | nindent 4 }}
+type: kubernetes.io/basic-auth
+data:
+  password: {{ .Values.objectStorageSecretOverride.password | default $oPassword | quote }}
+  username: {{ .Values.objectStorageSecretOverride.username | default $oUsername | quote }}
```
jujaga, line 65: We can drop the default generated fallback clauses ans just pipe quote the values here.

charts/coms/templates/secret.yaml
```diff
@@ -13,26 +21,62 @@
...
+  {{- if not .Values.config.releaseScoped }}
+  annotations:
+    "helm.sh/resource-policy": keep
+  {{- end }}
+  name: {{ $kSecretName }}
+  labels: {{ include "coms.labels" . | nindent 4 }}
+type: kubernetes.io/basic-auth
+data:
+  password: {{ .Values.keycloakSecretOverride.password | default $kPassword | quote }}
+  username: {{ .Values.keycloakSecretOverride.username | default $kUsername | quote }}
```
jujaga, line 81: We can drop the default generated fallback clauses ans just pipe quote the values here.

jujaga (changes requested):
Thanks for the general logic wrapper changes. I believe the and path is correct as we need to enforce the following:
1) If you already have a secret, do not touch it
2) If you don't have a secret, and you do have values specified, use them
The auto-generating behavior, while nice to have, is completely superfluous now with this logic flow and should be trimmed out.

charts/coms/templates/secret.yaml
cberg-aot: Refactored and gave it a sanity install for good measure.

jujaga (approved):

kamorel (approved):

PR #154, opened 2023-04-11
Feature/secret override

cberg-aot:
# Description
Refactored Secret objects to use stringData instead of Data as field.  This prevents base64 error when trying to pull data into the secret that may not be base64 (such as usernames).

## Types of changes
Refactor w/ encryption side effects (should make Secrets work as intended).

## Checklist

## Further comments
Tested with manual helm install, and NOT piping secrets to | base64.

This wasn't a tested workflow during the original PR (was feeding test data to base64, overlooking the usual flow of just placing stringData into the secret)


jujaga:
Superceded by #158

PR #191, opened 2023-08-02
Sync service for version, tags and metadata

TimCsaky:
# Description
syncs object data between S3 and COMS database.
A 'top-level' `syncJob()` function that in turn calls:
`syncObject()`, `syncVersions()`, `syncTags()` and `syncMetadata()`.

more details can be found in the overall sync service design:
https://gist.github.com/jujaga/29d74ac57bf1f9003e9420152393215b#sync-service-design

and the JIRA tickets:
- https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3253 (user story)
- https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3251
- https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3250

This should handle syncs between versions/un-versioned buckets, as well as delete markers.
Should also sync default bucket (`GET /sync`)
The following will be done in a separate PR:
- syncing all buckets in one shot
- Unit tests 

This PR also addresses the need to track `isLatest` in the database, and keep it accurate whenever new versions are created and deleted etc.

## Types of changes

## Checklist

### DIFF
--- a/app/src/services/version.js
+++ b/app/src/services/version.js
@@ -96,12 +96,11 @@ const service = {
         .returning('*');
 
       // if new version is latest, set all other versions to islatest: false
-      if(data.isLatest){
+      if (data.isLatest) {
         await Version.query(trx)
           .update({ 'isLatest': false, 'objectId': data.id })
           .whereNot({ 'id': response.id })
-          .andWhere('objectId', data.id)
-          .returning('*');
+          .andWhere('objectId', data.id);
       }
 
       if (!etrx) await trx.commit();
@@ -171,18 +170,12 @@ const service = {
       let response = undefined;
       if (s3VersionId) {
         response = await Version.query(trx)
-          .where({
-            s3VersionId: s3VersionId,
-            objectId: objectId
-          })
+          .where({ s3VersionId: s3VersionId, objectId: objectId })
           .first();
       }
       else if (versionId) {
         response = await Version.query(trx)
-          .where({
-            id: versionId,
-            objectId: objectId
-          })
+          .where({ id: versionId, objectId: objectId })
           .first();
       }
       else {
@@ -278,20 +271,14 @@ const service = {
       trx = etrx ? etrx : await Version.startTransaction();
       // update this version
       const updated = await Version.query(trx)
-        .update({
-          isLatest: isLatest,
-          objectId: objectId
-        })
+        .update({ isLatest: isLatest, objectId: objectId })
         .where({ id: id })
         .returning('*');
       // if we set this version with isLatest: true
       if (isLatest) {
         // set all other versions to islatest: false
         await Version.query(trx)
-          .update({
-            isLatest: false,
-            objectId: objectId
-          })
+          .update({ isLatest: false, objectId: objectId })
           .whereNot({ 'id': id })
           .andWhere('objectId', objectId)
           .returning('*');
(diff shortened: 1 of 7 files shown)


TimCsaky:
not sure why this didnt deploy. maybe because i had it as a draft initially (?)

app/src/services/version.js
```diff
@@ -268,27 +269,31 @@ const service = {
    * @param {string} options.id COMS uuid of a version
    * @param {string} options.objectId COMS uuid of an object
    * @param {boolean} options.isLatest isLatest value to set in db
-   * @param {string} [userId=SYSTEM_USER] The uuid of a user
    * @param {object} [etrx=undefined] An optional Objection Transaction object
    * @returns {object} Version model of updated version in db
    */
-  updateIsLatest: async ({ id, objectId, isLatest }, userId = undefined, etrx = undefined) => {
+  updateIsLatest: async ({ id, objectId, isLatest }, etrx = undefined) => {
```
jujaga, line 275: I'm not sure if we should be dropping userId attribution for this. Yes, this is technically a maintenance function for keeping things consistent, but the consistency trigger will still generally be invoked by some agent for some reason or another. I think we chatted about this previously as we were worried about update misattribution, so it's fine to leave it not updated I suppose, but removing it from the function signature I'm not certain of yet.

app/src/services/version.js
```diff
@@ -301,10 +306,16 @@ const service = {
...
-            .update({ 'isLatest': true, 'objectId': objectId, updatedBy: userId })
-            .where({ 'id': sq[0]?.id, 'objectId': objectId });
+            .update({
+              'isLatest': true,
+              'objectId': objectId,
+            })
+            .where({
+              'id': sq[0]?.id,
+              'objectId': objectId
+            });
```
jujaga, line 318: Do we need to be re-indenting the inner objects when they're just two attributes?

TimCsaky: good call. i've now condensed a few of these

app/src/services/version.js
```diff
@@ -91,17 +91,18 @@ const service = {
...
-        .andWhere('objectId', data.id)
-        .returning('*');
-
+      // if new version is latest, set all other versions to islatest: false
+      if(data.isLatest){
+        await Version.query(trx)
+          .update({ 'isLatest': false, 'objectId': data.id })
+          .whereNot({ 'id': response.id })
+          .andWhere('objectId', data.id)
+          .returning('*');
```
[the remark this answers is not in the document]
TimCsaky: not needed, removed

app/src/services/version.js
TimCsaky: I'm 100% sure either but i think this is best.
a few points to try and clarify:

- This function is typically called when a new version is created and you set it as latest.
the new version will have `createdBy: <user>`. 

- If the object is in an un-versioned bucket (there will only be one version), after an updateObject call, the single version does show `updatedBy: <user>`. 

- it is a maintainence process, I dont believe we need to update the updatedBy column again when we're just setting as latest in this separate process.

- we could just set `updatedBy: <user>` for the version that will become latest. but we definitely dont want to update that column for all the other versions we are setting as not latest.

- the object record (not version) also has createdBy and updatedBy attributes which are probably more useful.

norrisng-bc (approved):

app/src/services/version.js
jujaga: Let's remove the argument for now, but add in a TODO comment considering re-adding it back in down the line so that we don't forget about this consideration in the future. 👍

jujaga (approved):

PR #192, opened 2023-08-04
update db secret

BrandonSharratt:
deployment config expects app-db-username and app-db-password to be in the secret and only username and password exist

# Description

## Checklist


jujaga (approved):
Thanks for submitting a PR. Tentatively approving, but we're going to need to do a safety/sanity pass on infrastructure first before we can merge this in.

TimCsaky (approved):

PR #195, opened 2023-08-11
Bug/duplicate users

TimCsaky:
# Description

DB migration to remove dupes, add unique constraint on identityId and idp columns
conflict behaviour changed with simpler sql queries
wrapped whole login process in db transaction

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3297

## Types of changes

## Checklist

### DIFF
--- a/app/src/services/user.js
+++ b/app/src/services/user.js
@@ -68,30 +68,49 @@ const service = {
    * @throws The error encountered upon db transaction failure
    */
   createUser: async (data, etrx = undefined) => {
-    let trx;
+    let trx, response;
     try {
-      trx = etrx ? etrx : await User.startTransaction();
+      trx = etrx ? etrx : await User.startTransaction();      
+      console.log('createUser');
 
-      if (data.idp) {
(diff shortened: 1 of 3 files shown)


app/src/services/user.js
```diff
@@ -134,18 +149,26 @@ const service = {
...
+        
+      if (!oldUser) {
+        // Add user to system
+        response = await service.createUser(newUser, trx);
+      } else {
+        // Update user data if necessary
+        response = await service.updateUser(oldUser.userId, newUser, trx);
+      }
+    });
+    return response;
```
jujaga, line 171: Since we're using the trxWrapper, we should be taking advantage of returning the callback and returning the result of the trxWrapper instead of jumping scope blocks for our response. Ideally we'd want to do something like return await utils.trxWrapper(...) and the callback would return the service.create/update response instead of assigning it to a let value outside its scope.

jujaga (approved):

PR #198, opened 2023-08-18
Ensure coms-id tags buried in history are used when soft-deleted

jujaga:
# Description

This PR focuses on making sure that when objects are soft deleted that their previous coms-id values are still used during synchronization where appropriate.

- 50f62e4 Implement _deriveObjectId utility synchronize method
- 58cfed0 Ensure that isAtPath utility recognizes files at root level as valid
- aff9799 Drop partial implementation of all queryparam from syncBucket controller

[SHOWCASE-3317](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3317)

## Types of changes

## Checklist

### DIFF
--- a/app/src/services/sync.js
+++ b/app/src/services/sync.js
@@ -208,7 +208,7 @@ const service = {
       // await call to get versions from both the COMS db and S3
       const [comsVersionPromise, s3VersionPromise] = await Promise.allSettled([
         Version.query(trx).modify('filterObjectId', object.id).orderBy('createdAt', 'desc'),
-        storageService.listObjectVersion({ filePath: object.path, bucketId: object.bucketId })
+        storageService.listAllObjectVersions({ filePath: object.path, bucketId: object.bucketId })
       ]);
 
       // COMS versions
--- a/app/src/services/sync.js
+++ b/app/src/services/sync.js
@@ -14,41 +14,87 @@ const versionService = require('./version');
 
 /**
  * The Sync Service
- * sync data between object storage and the COMS database
+ * Synchronizes data between S3 object storage and the COMS database
  */
 const service = {
+  /**
+   * @function _deriveObjectId
+   * Checks an S3 Object for any previous `coms-id` tag traces and returns it if found.
+   * Otherwise it writes a new `coms-id` to the S3 Object if none were previously found.
+   * @param {object | boolean} s3Object The result of an S3 HeadObject operation
+   * @param {string} options.path String representing the canonical path for the specified object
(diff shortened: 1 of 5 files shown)


norrisng-bc (approved):

app/src/services/sync.js
```diff
@@ -81,7 +127,8 @@ const service = {
             response[0].versions.find(v => v.id === version.id).metadata = metadata;
           }
         }
-        log.verbose(`Finished syncing  ${path} in bucket ${bucketId}`, { function: 'syncJob', result: response });
+
+        log.verbose(`Finished syncing ${path} in bucket ${bucketId}`, { function: 'syncJob', result: response });
         return response;
```
TimCsaky, line 132: we don't actually expose this response to the api user.. but it seems like a nice thing to have (a nested array of object>version?meta/tags) for logging.

app/src/services/sync.js
```diff
@@ -91,95 +138,66 @@ const service = {
...
-   * @param {string} [options.bucketId] The uuid bucketId of object in sync job
+   * Synchronizes Object level data
+   * @param {string} options.path The path of object in sync job
+   * @param {string | null} options.bucketId The uuid bucketId of object in sync job
    * @param {string} [options.userId=SYSTEM_USER] The uuid of a user that created the sync job
    * @param {object} [etrx=undefined] An optional Objection Transaction object
-   * @returns {object} synced objects that exist in COMS and S3 eg:
-   * <ObjectModel> (synced object)
-   * or `undefined` (when object was pruned from COMS db after sync)
+   * @returns {Promise<Array<object | undefined>>} Either an array of synced objects or undefined
```
TimCsaky, line 148: i think this syncObject method returns just one js object 
`@returns {Promise<object | undefined>}`

app/src/services/sync.js
```diff
@@ -14,41 +14,87 @@ const versionService = require('./version');
...
+
+        // 1. Sync Object
+        const { newObject, ...object } = await service.syncObject({ path: path, bucketId: bucketId, userId: userId }, trx);
         response.push(object);
 
-        // 2. sync all versions of object
+        // 2. Sync Object Versions
         let versions = [];
-        // IF this is a new (or existing) object in COMS DB (from step 1)
-        // OR doing 'full' sync AND object wasn't deleted
```
TimCsaky, line 45: some kind of comments here explaining the logic are useful. we could add them later.
I think the current general behaviour is that **without** full mode, we only sync versions and meta/tags for **new** objects

TimCsaky (commented):
some good optimizations here

PR #200, opened 2023-08-23
Standardize sync calls to return modified flag and subject in object

jujaga:
# Description

Also modifies general function signatures to not be object option shaped as there are parameters that must be supplied in order to function. This also adds some missing etrx wrappers to existing service calls so that they can support transaction streaming.

[SHOWCASE-3313](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3313)

## Types of changes

## Checklist

## Further comments

Unit tests are known to be failing - this will be addressed in subsequent PR.

### DIFF
--- a/app/src/components/queueManager.js
+++ b/app/src/components/queueManager.js
@@ -82,15 +82,10 @@ class QueueManager {
 
         log.verbose(`Started processing job id ${job.id}`, { function: 'processNextJob', job: job });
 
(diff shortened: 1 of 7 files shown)


TimCsaky (approved):

TimCsaky:
Merged with only one review because we need to start testing with this and can fix bugs in subsequent PR's

PR #203, opened 2023-08-31
check for undefined TagSet

TimCsaky:
# Description

- handle TagSet when undefined in S3 GetObjectTaggingCommand() response.
steps to reproduce: 
mount add an empty Dell s3 bucket to coms
upload a new file using S3 Browser/cyberduck
call coms sync for the bucket.
console shows error, new object not synced

- other coms endpoints that use this command but because we add coms-id tag by default, issue did not present.

## Types of changes

## Checklist


jujaga:
Superceded by #204

PR #208, opened 2023-09-07
Validate default bucket configuration on startup

TimCsaky:
# Description

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3337
.. and some other changes required to make the default bucket feature optional.

- check default bucket doesnt also exist in database (prevents data integrity issues)
- simplify getBucket() function to always throw if no bucket found, update references to this function
- update github environment files with change of variables naming

## Types of changes

## Checklist


app/src/components/utils.js
```diff
@@ -61,42 +61,44 @@ const utils = {
...
-      // Function scoped import to avoid circular dependencies
-      const { bucketService } = require('../services');
-
-      try {
+  async getBucket(bucketId = undefined) {
+    try {
+      const data = { region: DEFAULTREGION };
+      if (bucketId) {
+        // Function scoped import to avoid circular dependencies
+        const { bucketService } = require('../services');
```
jujaga, line 76: I'm wondering if we should change this to be a deeper scoped import...
```js
const { read } = require('../services/bucket');
const bucketData = await read(bucketId);
```

jujaga (approved):
It's looking pretty good at a semantic glance. As we are shaking some things up with the helm chart env values, we are going to need to bump the helm chart version so that these changes will actually catch on CI/CD on merge.

norrisng-bc (approved):

app/src/components/utils.js
TimCsaky: seems like a good idea. tested and it works

wilwong89 (approved):

PR #224, opened 2023-10-26
Change SyncTags to only backfill coms-id tag for latest version

TimCsaky:
writing coms-id to tag to all versions in S3 alters Last Modified date for objects in Dell ECS, so we avoid soing this during a sync.

# Description

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3387

during sync of objects in Dell S3, the SyncTags() sub-routine adds coms-id tag which in turn was altering Last Modified property of all S3 versions. This is an unwanted side-affect that may confuse file owners looking at dates. (see JIRA ticket for logs)
Solution is to only add coms-id tag to latest version which does not alter Last Modified for any version.

## Checklist

### DIFF
--- a/app/src/services/sync.js
+++ b/app/src/services/sync.js
@@ -328,17 +328,25 @@ const service = {
       const comsTags = comsTagsForVersion[0]?.tagset ?? [];
       // S3 Tags
       const s3Tags = toLowerKeys(s3TagsForVersion?.TagSet ?? []);
-
-      // Ensure `coms-id` tag exists on this version in S3
-      if (s3Tags.length < 10 && !s3Tags.find(s3T => s3T.key === 'coms-id')) {
+      /**
+       * Add coms-id tag to latest version in S3 if not already present
+       * NOTE: For a sync job the _deriveObjectId() function will have already added
+       * the coms-id to latest version.
+       * TODO: check if this version is still also the latest on corresponding version in S3
+       */
+      if (comsVersion.isLatest && s3Tags.length < 10 && !s3Tags.find(s3T => s3T.key === 'coms-id')) {
+        /**
+         * NOTE: adding tags to a specified version (passing a `VersionId` parameter) will affect `Last Modified`
+         * attribute of multiple versions on some s3 storage providors including Dell ECS
+         */
         await storageService.putObjectTagging({
           filePath: path,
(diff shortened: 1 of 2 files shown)


app/src/services/sync.js
```diff
@@ -329,15 +329,18 @@ const service = {
...
+      if (version.isLatest && s3Tags.length < 10 && !s3Tags.find(s3T => s3T.key === 'coms-id')) {
+        // note: add tags to a specified version (passing a `VersionId` parameter) will affect `Last Modified`
+        // object attribute of versions on some s3 storage providors including Dell ECS
         await storageService.putObjectTagging({
           filePath: path,
           tags: (s3TagsForVersion?.TagSet ?? []).concat([{
             Key: 'coms-id',
             Value: comsVersion.objectId
           }]),
-          s3VersionId: comsVersion.s3VersionId,
```
jujaga, line 340: May be worth keeping this line in as a commented out value with the above note instead of outright removing it.

app/src/services/sync.js
```diff
@@ -329,15 +329,18 @@ const service = {
       // S3 Tags
       const s3Tags = toLowerKeys(s3TagsForVersion?.TagSet ?? []);
 
-      // Ensure `coms-id` tag exists on this version in S3
-      if (s3Tags.length < 10 && !s3Tags.find(s3T => s3T.key === 'coms-id')) {
+      // Add coms-id tag to latest version in S3 if not already present
+      // note: For a sync job the _deriveObjectId() function will have already added
+      // the coms-id to latest version.
+      if (version.isLatest && s3Tags.length < 10 && !s3Tags.find(s3T => s3T.key === 'coms-id')) {
```
jujaga, line 335: Just realizing that we're pulling isLatest data from COMS lineage, and not potentially from S3. We might need to do one extra storageService call to get the current version state and check the isLatest from there instead of a potentially derived coms one?

kyle1morel (approved):

app/src/services/sync.js
TimCsaky: as discussed, this is an edge case and acceptable risk we can consider fixing as required (for example, if we re-use SyncTags() outside of the complete object sync job.

jujaga (approved):

jatindersingh93 (approved):

wilwong89 (approved):

PR #237, opened 2023-12-27
Add pagination and sort order functionality to searchobject endpoint

jatindersingh93:
- Test using all five and individual query parameters (page=1&permissions=y&limit=3&sort=name&order=desc)

- Make sure key(X-Total-Rows) is present in header and updated with each request (with or without pagination)

- Test using Basic and JWT authentications

- Confirm all relevant permissions are being fetched

[SC3445](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3445)

# Description

## Checklist

### DIFF
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -929,14 +929,19 @@ const controller = {
         public: isTruthy(req.query.public),
         active: isTruthy(req.query.active),
         deleteMarker: isTruthy(req.query.deleteMarker),
-        latest: isTruthy(req.query.latest)
+        latest: isTruthy(req.query.latest),
+        permissions: isTruthy(req.query.permissions),
+        page: req.query.page,
+        limit: req.query.limit,
+        sort: req.query.sort,
+        order: req.query.order,
       };
       // if scoping to current user permissions on objects
       if (config.has('server.privacyMask')) {
         params.userId = await userService.getCurrentUserId(getCurrentIdentity(req.currentUser, SYSTEM_USER));
       }
       const response = await objectService.searchObjects(params);
-      res.status(200).json(response);
+      res.setHeader('X-Total-Rows', response.total).status(200).json(response.data);
     } catch (error) {
       next(error);
     }
--- a/app/src/db/models/tables/objectModel.js
+++ b/app/src/db/models/tables/objectModel.js
@@ -82,6 +82,7 @@ class ObjectModel extends Timestamps(Model) {
         if (value) {
           query
             .withGraphJoined('version')
+            .leftJoinRelated('version')
             .whereIn('version.id', builder => {
               builder.select('version.id')
                 .where('version.mimeType', 'ilike', `%${value}%`);
@@ -92,6 +93,7 @@ class ObjectModel extends Timestamps(Model) {
         if (value !== undefined) {
           query
             .withGraphJoined('version')
+            .leftJoinRelated('version')
             .where('version.deleteMarker', value);
         }
       },
@@ -145,6 +147,7 @@ class ObjectModel extends Timestamps(Model) {
         if (subqueries.length) {
           query
             .withGraphJoined('version')
+            .leftJoinRelated('version')
             .whereIn('version.id', builder => {
               builder.intersect(subqueries);
             });
@@ -156,6 +159,10 @@ class ObjectModel extends Timestamps(Model) {
       hasPermission(query, userId, permCode) {
         if (userId && permCode) {
           query
+            // withGraphFetched keep joining using default 'left join' operation,
+            // to fix default behavior we are adding extra joinOperation which seems to be working with
+            // corresponding JoinRelated
+            .withGraphFetched('[objectPermission, bucketPermission]', { joinOperation: 'fullOuterJoinRelated' })
             .fullOuterJoinRelated('[objectPermission, bucketPermission]')
             // wrap in WHERE to make contained clauses exclusive of root query
             .where(query => {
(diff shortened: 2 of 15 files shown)


app/src/controllers/object.js
```diff
@@ -929,14 +929,20 @@ const controller = {
         public: isTruthy(req.query.public),
         active: isTruthy(req.query.active),
         deleteMarker: isTruthy(req.query.deleteMarker),
-        latest: isTruthy(req.query.latest)
+        latest: isTruthy(req.query.latest),
+        page: req.query.page,
+        limit: req.query.limit,
+        sort: req.query.sort,
+        order: req.query.order,
+        permissions: isTruthy(req.query.permissions)
```
jujaga, line 937: Optional Style: have permissions come before the page/limit/sort/order as permissions is logical controller functionality - let's aim for pagination parameters to be conventionally the last to show up in a parameter set.

app/src/controllers/object.js
```diff
@@ -929,14 +929,20 @@ const controller = {
...
       };
       // if scoping to current user permissions on objects
       if (config.has('server.privacyMask')) {
         params.userId = await userService.getCurrentUserId(getCurrentIdentity(req.currentUser, SYSTEM_USER));
       }
       const response = await objectService.searchObjects(params);
-      res.status(200).json(response);
+      res.setHeader('X-Total-Rows', response.total);
+      res.status(200).json(response.data);
```
jujaga, line 945: Optional Consideration: We might be able to one-line this:
```js
res.setHeader('X-Total-Rows', response.total).status(200).json(response.data);
```
Not sure if it's better to do this though w.r.t. readability as they're functionally equivalent.

app/src/db/models/tables/objectModel.js
```diff
@@ -156,6 +159,7 @@ class ObjectModel extends Timestamps(Model) {
       hasPermission(query, userId, permCode) {
         if (userId && permCode) {
           query
+            .withGraphFetched('[objectPermission, bucketPermission]', { joinOperation: 'fullOuterJoinRelated' })
```
jujaga, line 162: Optional: Might be worth putting in a comment here to explain the nuance of what this graph fetch is doing. There is probably a significant amount of subtle behavior happening here with the extra joinOperation option.

app/src/db/models/tables/objectModel.js
```diff
@@ -175,7 +179,16 @@ class ObjectModel extends Timestamps(Model) {
                     });
                 });
             });
+        } else {
+          query
+            .withGraphFetched('objectPermission');
```
jujaga, line 184: Style: I would recommend one-lining this.

app/src/docs/v1.api-spec.yaml
```diff
@@ -401,6 +401,11 @@ paths:
         - $ref: "#/components/parameters/Query-Public"
         - $ref: "#/components/parameters/Query-MimeType"
         - $ref: "#/components/parameters/Query-Name"
+        - $ref: "#/components/parameters/Query-Permissions"
+        - $ref: "#/components/parameters/Query-Page"
+        - $ref: "#/components/parameters/Query-Limit"
+        - $ref: "#/components/parameters/Query-Sort"
+        - $ref: "#/components/parameters/Query-Order"
         - $ref: "#/components/parameters/Query-TagSet"
```
jujaga, line 409: Style: Let's preserve previous ordering by having the newly added parameters be after TagSet instead.

app/src/docs/v1.api-spec.yaml
```diff
@@ -1738,6 +1743,61 @@ components:
...
+      in: query
+      name: order
+      description: >-
+        Order in ascending or descending. Default to ascending order.
+      schema:
+        type: string
+        enum:
+          - asc
+          - desc
+        example: asc
```
jujaga, line 1800: Style: These parameter component definitions should stay alphabetized across the entire parameter subsection as the rest of the parameters were previously alphabetical order already.

app/src/services/object.js
```diff
@@ -117,11 +117,13 @@ const service = {
...
     let trx;
+    let response = [];
     try {
       trx = etrx ? etrx : await ObjectModel.startTransaction();
 
-      const response = await ObjectModel.query(trx)
-        .allowGraph('version')
+      response.data = await ObjectModel.query(trx)
+        .allowGraph('[objectPermission, version, bucketPermission]')
+        .groupBy('object.id')
```
jujaga, line 126: Optional: May be worth putting a comment about why groupBy was chosen over orderBy for future reference.

app/tests/unit/validators/object.spec.js
```diff
@@ -512,6 +512,47 @@ describe('searchObjects', () => {
...
+    });
+
+    describe('permissions', () => {
+      const permissions = query.keys.permissions;
+
+      it('is the expected schema', () => {
+        expect(permissions).toEqual(type.truthy.describe());
+      });
+    });
+
```
jujaga, line 555: These tests are pretty light and only check for absolute base typings. Would be ideal to introspect these objects a bit further and ensure that things like page minimum, limit minimum, sort bounded list and order enumeration definitions are found within the query.keys.* structure respectively. Validation unit tests are usually filled with `expect.arrayContaining(...)` and `expect objectContaining(...)` assertions to represent those kinds of bindings.

jujaga (commented):
Functionally looks good - great work! A few style considerations to take a look at, and ideally another pass on the validation unit tests for better functional definition coverage.

jujaga (approved):

kyle1morel (approved):

PR #247, opened 2024-02-07
Populate lastSyncedDate on object create/update

norrisng-bc:
# Description

`object.lastSyncedDate` is now populated on object create/update.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3506

## Types of changes

## Checklist

## Further comments

N/A


app/src/services/object.js
```diff
@@ -230,7 +231,8 @@ const service = {
         path: data.path,
         public: data.public,
         active: data.active,
-        updatedBy: data.userId ?? SYSTEM_USER
+        updatedBy: data.userId ?? SYSTEM_USER,
+        lastSyncedDate: new Date()
```
TimCsaky, line 235: i think this needs to be ` lastSyncedDate: data.lastSyncedDate`
and the value needs to come in as a parameter. if not provided it will be undefined and show as null in the db.
because there are places in the app where we call this update method but we're not actually syncing with all the data in S3.

TimCsaky (commented):
when someone runs the sync job.. we now always need to update the object.lastSyncedDate
this change should do it:
image

app/src/services/object.js
```diff
@@ -34,7 +34,8 @@ const service = {
         public: data.public,
         active: data.active,
         bucketId: data.bucketId,
-        createdBy: data.userId ?? SYSTEM_USER
+        createdBy: data.userId ?? SYSTEM_USER,
+        lastSyncedDate: new Date()
```
jujaga, line 38: This line is disingenuous. The intent of the lastSyncedDate is to record the last time an _explicit_ synchronization operation was applied to the object. When an object is freshly created via COMS, while it is technically synchronized with S3 as it was COMS that pushed up the new content, there was no actual synchronization operation performed. We should not be populating it here.

app/src/services/object.js
jujaga: This line is disingenuous and should not be included in the patch operation. The intent of the lastSyncedDate is to record the last time an _explicit_ synchronization operation was applied to the object. When an object is updated via COMS, you have no guarantee that COMS state is completely synchronized with S3. The most you can guarantee is that the latest new version uploaded exists in both COMS and S3. In the event S3 and COMS has already diverged prior to the update, (let's say 3 versions known in COMS, but S3 has 5 versions), after this operation, COMS still won't know about the other two versions without a formal sync. We should not be populating it here.

jujaga (commented):
The `lastSyncedDate` column should only be populated and modified via a /sync operation (stemming from syncBucket or syncObject for example). For standard object manipulations such as objectCreate and objectUpdate, whatever the lastSyncedDate value is must be preserved and not manipulated, as the column is meant to precisely report back when an object last had an _explicit_ synchronization operation applied. Object create and update are not able to safely assert that a complete synchronization has taken place as that is not what those endpoints are designed to do.

app/src/services/object.js
TimCsaky: after further discussion, i agree with Jeremy, the lastSyncDate should be null when an object is created or updated.
We only want to populate this field with the current timestamp when the syncObject() function is called (services/sync.js)

app/src/services/object.js
```diff
@@ -235,7 +234,7 @@ const service = {
         public: data.public,
         active: data.active,
         updatedBy: data.userId ?? SYSTEM_USER,
-        lastSyncedDate: new Date()
+        lastSyncedDate: data.lastSyncedDate ? data.lastSyncedDate : undefined
```
jujaga, line 237: This can be simplified to the following (assuming this doesn't accidentally nullify any values that are already there previously:
```js
lastSyncedDate: data.lastSyncedDate ?? undefined
```

TimCsaky: confirmed that updating with x: undefined is skipped

TimCsaky (approved):

app/src/services/object.js
```diff
@@ -232,7 +232,8 @@ const service = {
         path: data.path,
         public: data.public,
         active: data.active,
-        updatedBy: data.userId ?? SYSTEM_USER
+        updatedBy: data.userId ?? SYSTEM_USER,
+        lastSyncedDate: data.lastSyncedDate ?? undefined
```
jujaga, line 236: As I'm thinking about this further, I'm wondering if the ?? undefined is even necessary, as the only time this attribute would show up is if it was passed into the data argument explicitly; and by virtue of that happening, it'd already either be undefined, or contain a value of some form.

Extra note: We need to add a new missing JSDoc line above L221 to talk about the optional lastSyncedDate argument.

jujaga (approved):

app/src/services/object.js
norrisng-bc: I think I might've been thinking of other less dynamically typed languages when I wrote in `undefined` - like, the kind where trying to use it would trigger some kind of null pointer exception.

Is there any value in leaving it there as a documentation thing? Like, making it obvious that it's intended behaviour? 🤔

PR #249, opened 2024-02-26
Implement lastModifiedDate tracking support

jujaga:
# Description

This change begins to track the LastModified attribute in S3 versions. However, due to the HeadObjectCommand returning a 404/405 when asking about a delete marker, we will not be able to back-populate delete marker version records on the fly. The synchronization operation is not affected by this shortcoming as it uses a different pathway to surface this data.

[SHOWCASE-3505](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3505)

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -153,13 +153,16 @@ const controller = {
         // create or update version in DB (if a non-versioned object)
         const version = s3Response.VersionId ?
           await versionService.copy(
-            sourceS3VersionId, s3Response.VersionId, objId, s3Response.CopyObjectResult?.ETag, userId, trx
+            sourceS3VersionId, s3Response.VersionId, objId, s3Response.CopyObjectResult?.ETag,
+            s3Response.CopyObjectResult?.LastModified, userId, trx
           ) :
           await versionService.update({
             ...data,
             id: objId,
             etag: s3Response.CopyObjectResult?.ETag,
-            isLatest: true
+            isLatest: true,
+            lastModifiedDate: s3Response.CopyObjectResult?.LastModified
+              ? new Date(s3Response.CopyObjectResult?.LastModified).toISOString() : undefined
           }, userId, trx);
 
         // add metadata for version in DB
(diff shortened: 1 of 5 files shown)


app/src/services/version.js
```diff
@@ -90,7 +90,8 @@ const service = {
           createdBy: userId,
           deleteMarker: data.deleteMarker,
```
TimCsaky, line 91: I think you need to add lastModifiedDate to the db insert happening on line 48 of services/version.js when it does the copy process. This is invoked during the flow when updating/replacing metadata when the bucket doesnt have versioning.

TimCsaky: separate note.. i hope the lastModifiedDate for a version doesnt change when you edit tags on that version 😬

app/src/controllers/object.js
```diff
@@ -344,6 +346,11 @@ const controller = {
         }
       }
 
+      const s3Head = await storageService.headObject({
+        filePath: joinPath(bucketKey, req.currentUpload.filename),
+        bucketId: bucketId
+      }).catch(() => ({}));
```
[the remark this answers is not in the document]
jujaga: The intent here is to silently fail and continue in the event the last modified date is not acquirable. This is an expected scenario in the following situations:
- The latest object is a deleteMarker: in this situation, we can expect a 404 response, which if not caught will blow up the stack. We provide and return a silent empty object in order to ensure that the request proceeds even without the last modified date.
- There is an errant network issue that makes this call throw. As it is not on the critical path to be able to complete the operation at this stage, we do a silent catch and return in order to make sure the request code continues to execute.

app/src/controllers/object.js
```diff
@@ -1086,6 +1097,11 @@ const controller = {
...
```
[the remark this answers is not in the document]
jujaga: The intent here is to silently fail and continue in the event the last modified date is not acquirable. This is an expected scenario in the following situations:
- The latest object is a deleteMarker: in this situation, we can expect a 404 response, which if not caught will blow up the stack. We provide and return a silent empty object in order to ensure that the request proceeds even without the last modified date.
- There is an errant network issue that makes this call throw. As it is not on the critical path to be able to complete the operation at this stage, we do a silent catch and return in order to make sure the request code continues to execute.

app/src/services/version.js
jujaga: Good catch. Looking into adding and updating the copy service to account for the new value. Will investigate tag modification signatures.

kyle1morel (approved):

PR #260, opened 2024-05-02
Bug fix: Sync public flag on object

TimCsaky:
# Description

bug existed where public file would become non-public after sync.
logic we want is:
- if object is marked as public in s3 then mark as public in coms db.
- if s3 public attribute is false or undefined, keep coms public status.

## Checklist

### DIFF
--- a/app/src/services/sync.js
+++ b/app/src/services/sync.js
@@ -159,22 +159,18 @@ const service = {
 
         // Case: already synced - record & update public status as needed
         if (comsObject) {
-          if (s3Public === undefined || s3Public === comsObject.public) {
-            response = await objectService.update({
-              id: comsObject.id,
-              userId: userId,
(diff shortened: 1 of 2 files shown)


app/src/services/sync.js
```diff
@@ -159,22 +159,18 @@ const service = {
...
+          response = await objectService.update({
+            id: comsObject.id,
+            userId: userId,
+            path: comsObject.path,
+            // update if public in s3
+            public: s3Public ? s3Public : comsObject.public,
+            lastSyncedDate: new Date().toISOString()
+          }, trx);
+
+          // if public flag changed mark as modified
```
TimCsaky, line 172: this line os not important.
marking the object as `modified` does nothing. 
we had intended to use the modified flag to determine if versions needed syncing as well.. but we currently always syn the versions (so long as an object exists)

PR #289, opened 2024-11-23
syncObject: delete 'coms-id' tag on new file if id already in use

norrisng-bc:
# Description

When syncing, if a object not currently in COMS has a `coms-id` S3 tag with a value already in use inside COMS, delete the S3 tag and re-generate a new object ID before adding it to COMS.

This resolves a bug with renaming a file and syncing the folder it's in; COMS will mark the file as soft-deleted but fail to pick up the name change (thus resulting in a de-sync between COMS/S3). In BCBox, it will appear as if renaming the file and syncing it removes it.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3782

## Types of changes

## Checklist

### Motivation

*[Please see Jira ticket for a more detailed explanation](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3782)*

S3 doesn't support file renaming; any attempt to do so (e.g. in GeoDrive or any external S3 tool/browser) is actually 2 operations:
1. Create a copy of the existing file with the new name
2. Delete the existing file

In (1), the new copy will include the same S3 tags and metadata as the original.

In (2), if the bucket has file versioning enabled, then it is a soft deletion. For COMS, this means that the old file still fully exists in its database.

When syncing, COMS will detect (2) as a new file and will try to add it to COMS, using the same `coms-id` it sees in its S3 tags. Attempting to re-use the same id will result in database errors, since it's the primary key for a COMS object.

### Consequences on file version history

Unfortunately, the "new" file with the updated name will lose its version history, since it's held by the "old" copy. However, since it's only soft-deleted in a versioned bucket, the file history technically still exists; it's just that it's not associated with the renamed "active" file.

While it's probably technically possible (on *some* level) to have COMS "link" the renamed file to the old soft-deleted one, such that you'd get one continuous file history, this is probably overly complex for limited gain - the 2 files are  completely separate entities in S3, so you'd have to hack together some novel non-S3 concept to have them linked.

### DIFF
--- a/app/src/services/sync.js
+++ b/app/src/services/sync.js
@@ -21,23 +21,34 @@ const service = {
    * @function _deriveObjectId
    * Checks an S3 Object for any previous `coms-id` tag traces and returns it if found.
    * Otherwise it writes a new `coms-id` to the S3 Object if none were previously found.
+   * If the `coms-id` conflicts with an existing COMS object, it is replaced with a new one.
    * @param {object | boolean} s3Object The result of an S3 HeadObject operation
    * @param {string} path String representing the canonical path for the specified object
    * @param {string | null} bucketId uuid of bucket or `null` if syncing object in default bucket
-   * @returns {Promise<string>} Resolves to an existing objectId or creates a new one
+   * @returns {Promise<string>} Resolves to an existing objectId (if not already in COMS)
+   *                            or creates a new one
    */
   _deriveObjectId: async (s3Object, path, bucketId) => {
     let objId = uuidv4();
 
     if (typeof s3Object === 'object') { // If regular S3 Object
-      const TagSet = await storageService.getObjectTagging({ filePath: path, bucketId: bucketId })
+      let TagSet = await storageService.getObjectTagging({ filePath: path, bucketId: bucketId })
         .then(result => result.TagSet ?? []);
+
       const s3ObjectComsId = TagSet.find(obj => (obj.Key === 'coms-id'))?.Value;
 
-      if (s3ObjectComsId && uuidValidate(s3ObjectComsId)) {
+      if (s3ObjectComsId
+        && uuidValidate(s3ObjectComsId)
+        && !(await objectService.exists(s3ObjectComsId))) {
+        // re-use existing coms-id (if no conflict)
         objId = s3ObjectComsId;
-      } else { // Update S3 Object if there is still remaining space in the TagSet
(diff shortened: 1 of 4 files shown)


app/src/services/sync.js
```diff
@@ -175,7 +175,43 @@ const service = {
...
-          const objId = await service._deriveObjectId(s3Object, path, bucketId);
+          let objId = await service._deriveObjectId(s3Object, path, bucketId);
+          let existingComsId;
+          try {
+            existingComsId = await objectService.read(objId);
+          } catch {
+            // Derived object id not currently in use by COMS
+          }
+
+          // Object id in 'coms-id' S3 tag already in use in COMS:
```
TimCsaky, line 186: maybe tidy up the comments?

app/src/services/sync.js
```diff
@@ -21,23 +21,34 @@ const service = {
...
+      if (s3ObjectComsId
+        && uuidValidate(s3ObjectComsId)
+        && !(await objectService.exists(s3ObjectComsId))) {
+        // re-use existing coms-id (if no conflict)
         objId = s3ObjectComsId;
-      } else { // Update S3 Object if there is still remaining space in the TagSet
-        if (TagSet.length < 10) { // putObjectTagging replaces S3 tags so new TagSet must contain existing values
+      } else {
+        // remove `coms-id` tag since it conflicts with an existing COMS object
+        TagSet.filter(x => x.Key != 'coms-id');
```
TimCsaky, line 47: i think `TagSet.filter()` creates a copy but wont change the value of `TagSet`. you need to re-assign it. (eg: `const filtered = Tagse..`) and then use that const in the code below.
or you could change TagSet with: `TagSet.splice(TagSet.findIndex(e => e.Key=== "coms-id"),1);`

app/src/services/sync.js
norrisng-bc: Implementation has since changed - previously this was all in `syncObject()`, it's now mainly handled in `_deriveObjectId()`

app/src/services/sync.js
norrisng-bc: Good catch - I've tweaked the`TagSet` declaration a little further up (it's now a variable instead of a constant) and made this line assign the filtered results (instead of just `filter()`'ing it and calling it a day).

TimCsaky (approved):

PR #290, opened 2024-12-06
Sync versions object during object create

TimCsaky:
Sync versions object during object create

### DIFF
--- a/app/src/controllers/object.js
+++ b/app/src/controllers/object.js
@@ -34,6 +34,7 @@ const {
   metadataService,
   objectService,
   storageService,
+  syncService,
   tagService,
   userService,
   versionService,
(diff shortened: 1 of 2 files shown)


app/src/controllers/object.js
```diff
@@ -297,20 +298,32 @@ const controller = {
...
-          throw new Problem(409, {
-            detail: 'Bucket already contains object',
-            instance: req.originalUrl
+
+          // sync existing file versions and get COMS objectId
+          const objectId =
+            await syncService.syncJob(joinPath(bucketKey, req.currentUpload.filename), bucketId, true, userId);
+
+          throw new Problem(409, 'Bucket already contains a soft-deleted version of this object', req.originalUrl, {
+            detail: 'Bucket already contains a soft-deleted version of this object',
```
norrisng-bc, line 323: IMO I'd suggest in the `detail` to upload a new version to the existing object instead, but I don't think our other `detail` messages (presumably by design) have this kind of `{error} + {suggestion}` format. Just a thought 🤷

PR #294, opened 2025-02-07
Sc3828

jatindersingh93:
- provide basic auth using accessKeyId & secretAccessKey
- Capture update logs
- Updates API docs
- Unit testing
- update pipelines to enable `S3ACCESSMODE_ENABLED true` by default

# Description

## Checklist

### DIFF
--- a/app/config/custom-environment-variables.json
+++ b/app/config/custom-environment-variables.json
@@ -1,6 +1,7 @@
 {
   "basicAuth": {
     "enabled": "BASICAUTH_ENABLED",
+    "s3AccessMode": "S3ACCESSMODE_ENABLED",
     "password": "BASICAUTH_PASSWORD",
     "username": "BASICAUTH_USERNAME"
   },
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -71,6 +71,9 @@ paths:
         generating a new bucket record. This endpoint can be used to grant the current user
         permission to upload to a new or existing bucket.
       operationId: createBucket
+      parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
       tags:
         - Bucket
       requestBody:
@@ -102,6 +105,8 @@ paths:
       tags:
         - Bucket
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Query-BucketId"
         - $ref: "#/components/parameters/Query-Active"
         - $ref: "#/components/parameters/Query-Key"
@@ -129,6 +134,8 @@ paths:
       tags:
         - Bucket
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Path-BucketId"
       responses:
         "204":
@@ -144,6 +151,8 @@ paths:
       tags:
         - Bucket
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Path-BucketId"
       responses:
         "200":
@@ -170,6 +179,8 @@ paths:
               items:
                 $ref: "#/components/schemas/Request-UpdateBucket"
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Path-BucketId"
       responses:
         "200":
@@ -194,6 +205,8 @@ paths:
       tags:
         - Bucket
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Path-BucketId"
       responses:
         "204":
@@ -215,6 +228,8 @@ paths:
       tags:
         - Bucket
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Path-BucketId"
       requestBody:
         content:
@@ -279,6 +294,8 @@ paths:
         - Sync
       operationId: syncBucket
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Path-BucketId"
       responses:
         "202":
@@ -303,6 +320,8 @@ paths:
         - Metadata
       parameters:
         - $ref: "#/components/parameters/Header-Metadata"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
       responses:
         "200":
           description: Returns an array of matching key/value pairs.
@@ -332,9 +351,11 @@ paths:
       tags:
         - Object
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
         - $ref: "#/components/parameters/Header-ContentDisposition"
         - $ref: "#/components/parameters/Header-ContentLength"
         - $ref: "#/components/parameters/Header-ContentType"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Header-Metadata"
         - $ref: "#/components/parameters/Query-TagSet"
         - in: query
@@ -403,6 +424,8 @@ paths:
       tags:
         - Object
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Header-Metadata"
         - $ref: "#/components/parameters/Query-ObjectId"
         - $ref: "#/components/parameters/Query-BucketId"
@@ -469,6 +492,8 @@ paths:
       tags:
         - Object
       parameters:
+        - $ref: "#/components/parameters/Header-S3AccessMode-Bucket"
+        - $ref: "#/components/parameters/Header-S3AccessMode-Endpoint"
         - $ref: "#/components/parameters/Path-ObjectId"
         - $ref: "#/components/parameters/Query-Download"
         - $ref: "#/components/parameters/Query-ExpiresIn"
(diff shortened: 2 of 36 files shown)


app/src/middleware/authentication.js
```diff
@@ -95,7 +127,7 @@ const currentUser = async (req, res, next) => {
   req.currentUser = Object.freeze(currentUser);
 
   // Continue middleware stack based on detected AuthType
-  if (currentUser.authType === AuthType.BASIC) {
+  if (currentUser.authType === AuthType.BASIC && getConfigBoolean('basicAuth.enabled')) {
     _checkBasicAuth(req, res, next);
   }
```
TimCsaky, line 132: i wonder if lines 130-132 should be moved to line 62.. to make it simpler

app/src/services/bucket.js
```diff
@@ -7,6 +7,38 @@ const { Bucket } = require('../db/models');
  * The Bucket DB Service
  */
 const service = {
+
+  /**
+   * @function checkBucketBasicAccess
+   * Grants a user provided permissions to the bucket if the data precisely matches
+   * accessKeyId, bucketId, bucket and endpoint values.
+   * @param {string} data.bucketId The S3 bucket access key id
+   * @param {string} data.bucket The S3 bucket identifier
```
TimCsaky, line 16:
```js
   * @param {string} data.bucketId The COMS guid for the bucket
   * @param {string} data.bucket The S3 bucket name
   ```

File: app/src/middleware/authorization.js [hunk 3]
```diff
@@ -44,6 +48,59 @@ const _checkPermission = async ({ currentObject, currentUser, params }, permissi
...
+ * @param {function} next The next callback function
+ * @returns {function} Express middleware function
+ * @throws The error encountered upon failure
+ * @returns
+ */
+const checkS3BasicAccess = async (req, _res, next) => {
+  const authType = req.currentUser ? req.currentUser.authType : undefined;
+  const bucketSettings = req.currentUser?.bucketSettings ? req.currentUser.bucketSettings : undefined;
+  let bucketIds = mixedQueryToArray(req.query.bucketId) || req.body.bucketId;
+  const objIds = mixedQueryToArray(req.query.objectId) || req.body.objectId;
```
TimCsaky, line 65: `const objIds = req.params.objectId`

app/config/custom-environment-variables.json
```diff
@@ -1,6 +1,7 @@
 {
   "basicAuth": {
     "enabled": "BASICAUTH_ENABLED",
+    "s3AccessMode": "BASICAUTH_ENABLED",
```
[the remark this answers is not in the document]
TimCsaky: create new environment variable

app/src/routes/v1/bucket.js
```diff
@@ -35,7 +36,8 @@ router.get('/', bucketValidator.searchBuckets, (req, res, next) => {
 });
 
 /** Updates a bucket */
-router.patch('/:bucketId', express.json(), bucketValidator.updateBucket, hasPermission(Permissions.MANAGE),
```
TimCsaky, line 38: I think we need to add `checkS3BasicAccess` middleware to nearly all the routes.
except for `PUT /bucket`

TimCsaky: for the 'search' endpoints where bucketId or objectId params are not required and wont be checked by checkS3BasicAccess..
but as discussed, we could force client to provide bucketId or objectId

but I suspect in some cases we will need to modify the sql queries.
see changes here: https://github.com/bcgov/common-object-management-service/compare/SC3828...s3-rev?expand=1

TimCsaky (commented):
after finally getting my hear around the logic.. it feels like youve choosen the right design.. looking good so far.
I only tested this with the search objects (GET /object) endpoint .. more testing required.

app/src/middleware/authentication.js
```diff
@@ -55,8 +56,34 @@ const currentUser = async (req, res, next) => {
...
+            secretAccessKey: secretAccessKey,
+          };
+          const bucketHeader = await storageService.headBucket(bucketSettings);
+
+          if (bucketHeader?.$metadata?.httpStatusCode === 200) {
+            currentUser.authType = AuthType.BASIC;
+            delete bucketSettings.secretAccessKey;
+            currentUser.bucketSettings = bucketSettings;
+          }
+        } catch (err) {
```
TimCsaky, line 83: i think lines 76-82 can be replaced with:
```js
          await storageService.headBucket(bucketSettings);
          delete bucketSettings.secretAccessKey;
          currentUser.bucketSettings = bucketSettings;
```

app/src/middleware/authorization.js
```diff
@@ -44,6 +48,54 @@ const _checkPermission = async ({ currentObject, currentUser, params }, permissi
...
+        bucketId: bucketIds,
+        bucket: bucketSettings.bucket,
+        endpoint: bucketSettings.endpoint,
+        accessKeyId: bucketSettings.accessKeyId,
+      };
+      const buckets = await bucketService.checkBucketBasicAccess(bucketData);
+
+      if (buckets.length === 0) {
+        return next(new Problem(403, {
+          detail: 'User lacks permission to complete this action',
```
TimCsaky, line 85: maybe replace with 
`          detail: 'The provided S3 credentials do not authorize this operation',
`
user doesnt really make sense in s3 access mode

app/src/middleware/authorization.js
```diff
@@ -44,6 +48,54 @@ const _checkPermission = async ({ currentObject, currentUser, params }, permissi
...
+          instance: req.originalUrl
+        }));
+      } else {
+        //bucketId params will be overwritten with passed or valid access bucketId.
```
TimCsaky, line 89: for some reason for `GET /bucket/:bucketId` (readBucket) endpoint, the `req.params` and `req.query` json are showing up as empty `{}`

TimCsaky: results in returning a bucket that doesnt match credentials in header

app/src/services/bucket.js
```diff
@@ -7,6 +7,38 @@ const { Bucket } = require('../db/models');
...
+   * accessKeyId, bucketId, bucket and endpoint values.
+   * @param {string} data.bucketId The COMS guid for the bucket
+   * @param {string} data.bucket The S3 bucket name
+   * @param {string} data.endpoint The S3 bucket endpoint
+   * @param {string} data.accessKeyId The S3 bucket secret access key
+   * @param {object} [etrx=undefined] An optional Objection Transaction object
+   * @returns {Promise<object>} The result will include all bucketIds that pass the validation
+   * @throws The error encountered upon db transaction failure
+   */
+  checkBucketBasicAccess: async (data, etrx = undefined) => {
```
TimCsaky, line 23: maybe consider using the existing `bucketService.readUnique()` function to reduce dupication
the result could be modified after back in the authorization.js

TimCsaky: actually maybe not. nothing wrong with having a dedicated function for clarity

TimCsaky (commented):
Looking good. your design is working well.. not sure we had a better option. great job.

app/src/services/bucket.js
```diff
@@ -7,6 +7,38 @@ const { Bucket } = require('../db/models');
...
+    let trx;
+    try {
+      trx = etrx ? etrx : await Bucket.startTransaction();
+
+      // Get existing record from DB
+      const buckets = await Bucket.query()
+        .modify('filterBucketIds', data.bucketId)
+        .where('bucket', data.bucket)
+        .where('endpoint', data.endpoint)
+        .where('accessKeyId', data.accessKeyId);
```
norrisng-bc, line 33: Isn't `accessKeyId` the username and `secretAccessKey` the password? That is, the "access key identifier" and "secret access key" in BCBox/S3 terms.

**Shouldn't we also verify the username ("access key identifier")** and not just the "secret access key"?

In any case, I think the param naming is a little confusing here, since "access key ID" implies the S3 bucket "username".

TimCsaky (commented):
another thing we need to add:
for the createObject, creatBucket and createChildBucket operations, we automatically give the user all permissions on the resource if userId != SYSTEM_USER.
see:
- https://github.com/bcgov/common-object-management-service/blob/cd5e836b474ecb985f38709629de37746e2f06a4/app/src/services/bucket.js#L98
- https://github.com/bcgov/common-object-management-service/blob/cd5e836b474ecb985f38709629de37746e2f06a4/app/src/services/object.js#L42

For S3 Access Mode we also cannot give current user permissions. 
not sure the best way to handle this. eg leave userId as SYSTEM_USER or somehow skip permissions insert if using S3AccessMode.

app/src/docs/v1.api-spec.yaml
```diff
@@ -2653,7 +2735,7 @@ components:
             status:
               example: 403
             detail:
-              example: User lacks permission to complete this action
+              example: The provided S3 credentials do not authorize this operation
```
TimCsaky, line 2738: not sure if we want this change here. i thought I took it out

app/src/middleware/authentication.js
TimCsaky: not sure if i mistakenly put some of this back during my merge

TimCsaky (commented):
not sure if i reverted any of your changes with my merge.
testing round soon

app/src/middleware/authorization.js
```diff
@@ -44,6 +52,56 @@ const _checkPermission = async ({ currentObject, currentUser, params }, permissi
...
+        bucketIds = objects.data.map(i => i.bucketId);
+      }
+    }
+
+    // filter request by buckets matching provided credentials
+    try {
+      const bucketData = {
+        bucketId: bucketIds,
+        bucket: bucketSettings.bucket,
+        endpoint: bucketSettings.endpoint,
```
norrisng-bc, line 90: If the endpoint as stored in the DB does not have a trailing slash but `x-amz-endpoint` does, I can retrieve *all* COMS buckets across all users.

e.g. `bucket.endpoint` (in the database) is `http://example.com`, but in the API call I have the header `x-amz-endpoint: http://example.com/`.

`utils.js` has a `stripDelimit()` helper function that should fix things:
```js
        endpoint: stripDelimit(bucketSettings.endpoint)
```

norrisng-bc:
Not related to the issue at hand, but could you slipstream a quick edit here? 

Specifically, changing `name: name` to `name: displayName` - COMS recognizes the latter but not the former.

https://github.com/bcgov/common-object-management-service/blob/5005973e21315ec4dbf8fe839b66af884ccff81b/app/src/docs/v1.api-spec.yaml#L1821

norrisng-bc:
> > > another thing we need to add: for the createObject, creatBucket and createChildBucket operations, we automatically give the user all permissions on the resource if userId != SYSTEM_USER. see:
> > > 
> > > * https://github.com/bcgov/common-object-management-service/blob/cd5e836b474ecb985f38709629de37746e2f06a4/app/src/services/bucket.js#L98
> > > * https://github.com/bcgov/common-object-management-service/blob/cd5e836b474ecb985f38709629de37746e2f06a4/app/src/services/object.js#L42
> > > 
> > > For S3 Access Mode we also cannot give current user permissions. not sure the best way to handle this. eg leave userId as SYSTEM_USER or somehow skip permissions insert if using S3AccessMode.
> 
> Currently, no permissions are being added when an object or bucket is created using s3AccessMode. We've decided to exclude permissions at this stage of the implementation.

I think we should add a note in the docs mentioning that you have to explicitly `PUT /permission/bucket/{bucketId}` after creating said bucket; otherwise, you've effectively created an inaccessible bucket (okay, you can access it with S3 credentials, but at that point you might as well directly access it over S3).

norrisng-bc:
https://github.com/bcgov/common-object-management-service/blob/5d97e3ec2ad8d582cb3d4bb96a9847dc64b60aec/app/src/docs/v1.api-spec.yaml#L218-L226

Would be helpful to mention that valid S3 credentials for the parent bucket are an acceptable alternative to OIDC Auth `MANAGE` permissions on the parent bucket.

TimCsaky (approved):
nice work!

PR #301, opened 2025-04-14
Add lastModifiedDate from version to the object response

jatindersingh93:
- Adding 'lastModifiedDate' to the search result, which can replace 'updated date' in the object list view.

# Description

## Checklist

### DIFF
--- a/app/src/services/object.js
+++ b/app/src/services/object.js
@@ -174,6 +174,11 @@ const service = {
                     .map(o => o.permCode);
                 }
               }
+              if (Array.isArray(version) && version.length > 0) {
+                object.lastModifiedDate = version.reduce((latest, v) => {
+                  return v.lastModifiedDate > latest ? v.lastModifiedDate : latest;
+                }, version[0].lastModifiedDate);
+              }
               return object;
             }).filter(x => x) // Drop empty row results from the array set
           );
--- a/app/src/services/object.js
+++ b/app/src/services/object.js
@@ -175,10 +175,12 @@ const service = {
                 }
               }
               if (Array.isArray(version) && version.length > 0) {
-                object.lastModifiedDate = version.reduce((latest, v) => {
-                  return v.lastModifiedDate > latest ? v.lastModifiedDate : latest;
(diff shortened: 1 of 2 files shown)


app/src/services/object.js
```diff
@@ -174,6 +174,11 @@ const service = {
                     .map(o => o.permCode);
                 }
               }
+              if (Array.isArray(version) && version.length > 0) {
+                object.lastModifiedDate = version.reduce((latest, v) => {
+                  return v.lastModifiedDate > latest ? v.lastModifiedDate : latest;
+                }, version[0].lastModifiedDate);
+              }
```
TimCsaky, line 181: i think if user does object search and filters on any version attributes, tag or metadata, the lastModifiedDate will be latest of those versions, not the necessarily the lastModifiedDate for the latest version of the object.

wonder if we would instead we should do (around line 155)
```js
.joinRelated('version', { alias: 'v' })
.where(v.isLatest, true)
...
# and then add that to the final object around 177
```

TimCsaky:
we can also include `lastodifiedDate` in the response from createObject and updateObject.
in controllers/object.js.. for updateObject[here](https://github.com/bcgov/common-object-management-service/blob/91569921c16be13e614214d02aa801071a62d526/app/src/controllers/object.js#L1214) and for createObject [here](https://github.com/bcgov/common-object-management-service/blob/91569921c16be13e614214d02aa801071a62d526/app/src/controllers/object.js#L382) add this line:
`object.lastModifiedDate = version.lastModifiedDate`

TimCsaky (approved):

norrisng-bc (approved):

norrisng-bc:
https://github.com/bcgov/common-object-management-service/blob/fd4d5c53461d50b3bea6fbe4f634bdf9462a8445/app/src/docs/v1.api-spec.yaml#L2410-L2414

I'd probably also clarify in the `description` that it's referring to a COMS record - specifically, operations done via COMS as well as when an external change was picked up during a sync.

PR #302, opened 2025-04-18
make public using bucket policy

TimCsaky:
# Description

WIP
using bucket Policies instead of ACL's is more effective.
Policies can also be applied to 'subfolders' (prefixes).
This pr swaps out the use of acl with policy for making a single obect public/private.

## Checklist

### DIFF
--- a/app/src/services/storage.js
+++ b/app/src/services/storage.js
@@ -12,6 +12,9 @@ const {
   ListObjectsV2Command,
   ListObjectVersionsCommand,
   PutObjectAclCommand,
+  GetBucketPolicyCommand,
+  PutBucketPolicyCommand,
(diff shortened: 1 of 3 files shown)


app/src/services/storage.js
```diff
@@ -497,6 +500,92 @@ const objectStorageService = {
...
+    }
+
+    // else making public
+    else {
+      // add policy
+      const policy = {
+        'Version': '2012-10-17',
+        'Statement': [
+          {
+            'Sid': 'coms::' + data.bucket + '/' + filePath,
```
TimCsaky, line 574: we can give our policies a distinct name (`coms::<file path>`) making it easier to remove them later without altering any other policies set externally

PR #306, opened 2025-05-07
Block unsecified idp's

TimCsaky:
# Description

Block idp's not specified in environment variable

## Types of changes

## Checklist

### DIFF
--- a/app/src/middleware/authentication.js
+++ b/app/src/middleware/authentication.js
@@ -116,7 +116,15 @@ const currentUser = async (req, res, next) => {
           throw new Error('OIDC environment variable KC_PUBLICKEY or keycloak.publicKey must be defined');
         }
 
-        if (isValid) {
+        // ensure token contains supported idp claim
+        const decodedToken = jwt.decode(bearerToken);
+        let idpSupported = true;
+        if (config.has('keycloak.identityKey')) {
+          const supportedIdentityKeys = config.get('keycloak.identityKey').split(',');
+          idpSupported = supportedIdentityKeys.some(k => Object.keys(decodedToken).includes(k));
+        }
+
+        if (isValid && idpSupported) {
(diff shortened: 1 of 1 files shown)


app/src/middleware/authentication.js
```diff
@@ -116,7 +116,15 @@ const currentUser = async (req, res, next) => {
           throw new Error('OIDC environment variable KC_PUBLICKEY or keycloak.publicKey must be defined');
         }
 
-        if (isValid) {
+        // ensure token contains supported idp claim
+        const decodedToken = jwt.decode(bearerToken);
+        let idpSupported = true;
+        if (config.has('keycloak.identityKey')) {
+          const supportedIdentityKeys = config.get('keycloak.identityKey').split(',');
```
norrisng-bc, line 123: IMO checking the `identity_provider` claim is a bit neater, but:
1. This would require a new environment variable (or that we re-purpose `keycloak.identityKey`).
2. This doesn't quite work for BCSC in the future because it doesn't even have an `identity_provider` claim.

I *might* be thinking a bit too far ahead with (2) though.

app/src/middleware/authentication.js
```diff
@@ -116,7 +116,15 @@ const currentUser = async (req, res, next) => {
           throw new Error('OIDC environment variable KC_PUBLICKEY or keycloak.publicKey must be defined');
         }
 
-        if (isValid) {
+        // ensure token contains supported idp claim
+        const decodedToken = jwt.decode(bearerToken);
```
norrisng-bc, line 120: I think you can reuse `isValid` from further up (albeit with a possible rename for readability) - it already returns the decoded token as long as the JWT is valid:

https://github.com/bcgov/common-object-management-service/blob/5224413f96dafd41a497a0f78679cb56b61d9d22/app/src/middleware/authentication.js#L112-L114

PR #309, opened 2025-06-21
Feature: optionally cascade add/remove bucket permissions operations

norrisng-bc:
# Description
This PR introduces an optional boolean URL parameter, `recursive`, to the following endpoints:
 * `PUT /permission/bucket/:bucketId` (add bucket permissions)
 * `DELETE /permission/bucket/:bucketId` (remove bucket permissions)

If `true`, the operation is also applied to all child buckets where the current user (i.e. the one making the API call) has `MANAGE` permissions. If the call is being made with basic or S3 auth, the operation is applied to *all* child buckets.

`recursive` defaults to `false`.

Some minor formatting tweaks were also made to the existing OpenAPI spec - specifically, standardizing the formatting for permCodes and URL params to `code blocks`.

## Types of changes

## Checklist

## Further comments

Bucket permissions are essentially encoded as `(bucketId, userId, permCode)` tuples in the the `bucket_permission` database table. 

Because of this, some quirks occur when using basic or S3 auth, caused by `req.currentUser` being `SYSTEM_USER` in such cases.

The `bucket_permission` table will not have any entries where `userId == SYSTEM_USER` (since `SYSTEM_USER` by definition has implicit access to *everything*), so there are places where it is necessary to bypass matching on `userId` when searching the `bucket_permission` table.

This bypassing was added to the existing `bucketService.searchChildBuckets()` function, and is also part of the new code for this new feature.

### DIFF
--- a/app/src/controllers/bucketPermission.js
+++ b/app/src/controllers/bucketPermission.js
@@ -1,4 +1,6 @@
+const { Permissions } = require('../components/constants');
 const errorToProblem = require('../components/errorToProblem');
+const utils = require('../db/models/utils');
 const {
   addDashesToUuid,
   mixedQueryToArray,
@@ -7,7 +9,7 @@ const {
   isTruthy
 } = require('../components/utils');
 const { NIL: SYSTEM_USER } = require('uuid');
-const { bucketPermissionService, userService } = require('../services');
+const { bucketPermissionService, userService, bucketService } = require('../services');
 
 const SERVICE = 'BucketPermissionService';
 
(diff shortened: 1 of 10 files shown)


app/src/controllers/bucketPermission.js
```diff
@@ -112,14 +159,36 @@ const controller = {
...
+
+        const allBuckets = [parentBucket, ...childBuckets];
+
+        const responses = await Promise.all(
+          allBuckets.map(b =>
+            // Only remove permissions from child buckets that currentUser can MANAGE
+            bucketPermissionService.removePermissions(b.bucketId, userIds, permissions)
+          )
+        );
+        res.status(200).json(responses);
```
TimCsaky, line 180: to keep the response in the same shape we can do this:
`res.status(201).json(responses.flat());`

note: when doing recursive the response array includes all the recursive new/deleted perm records.. which I think is what we want.

app/src/controllers/bucketPermission.js
```diff
@@ -89,11 +117,30 @@ const controller = {
...
+
+        const allBuckets = [parentBucket, ...childBuckets];
+
+        const responses = await Promise.all(
+          allBuckets.map(b =>
+            // Only apply permissions to child buckets that currentUser can MANAGE
+            bucketPermissionService.addPermissions(b.bucketId, req.body, currUserId)
+          )
+        );
+        res.status(201).json(responses);
```
TimCsaky, line 137: `res.status(201).json(responses.flat());`

TimCsaky (approved):

PR #320, opened 2026-05-27
AzureIDIR (IDIR MFA) support

norrisng-bc:
# Description

This PR introduces full support for IDIR MFA (a.k.a. `azureidir`) as an identity provider, alongside the existing SiteMinder IDIR (or `idir`). 

`azureidir` was already implicitly supported, since `azureidir`-based JWTs also come from the Pathfinder SSO common Keycloak realm (this allows clients to supply their own JWTs). This PR "merges" the two identities, in the sense that whenever `azureidir` is present in the incoming JWT's `identity_provider` or in the query parameters, it is overwritten/stored as `idir` inside COMS.

Additionally, the behaviour in #311 has now been extended to also include IDP-based permissions. This allows such permissions to be inherited when a child bucket is created.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3713
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-4364

## Types of changes

## Checklist

## Further comments

After discussion, it was agreed that overwriting the `identity_provider` field at the auth middleware layer would be the simplest way to merge `idir` and `azureidir`. While this certainly isn't the cleanest, other methods would've entailed modifying significantly more code, edge case handling, updating tests etc.

### DIFF
--- a/app/src/services/user.js
+++ b/app/src/services/user.js
@@ -158,7 +158,7 @@ const service = {
     return await utils.trxWrapper(async (trx) => {
       // check if user exists in db
       const oldUser = await User.query(trx)
-        .where({ 'identityId': newUser.identityId, idp: newUser.idp })
+        .where({ 'identityId': newUser.identityId })
         .first();
 
       if (!oldUser) {
--- a/app/src/controllers/bucket.js
+++ b/app/src/controllers/bucket.js
@@ -17,7 +17,13 @@ const {
 } = require('../components/utils');
 const { redactSecrets } = require('../db/models/utils');
 
-const { bucketService, storageService, userService, bucketPermissionService } = require('../services');
+const {
+  bucketService,
+  storageService,
+  userService,
+  bucketPermissionService,
+  bucketIdpPermissionService
+} = require('../services');
 
 const SERVICE = 'BucketService';
 const secretFields = ['accessKeyId', 'secretAccessKey'];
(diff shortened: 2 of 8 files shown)


app/src/components/constants.js
```diff
@@ -29,7 +29,8 @@ module.exports = Object.freeze({
    * are given additional permissions.
   */
   ElevatedIdps: [
-    'idir'
+    'idir',
+    'azureidir'
```
TimCsaky: dont think we need this if idp is always `idir` (?)

app/src/services/user.js
TimCsaky, line 169: I thought we were going to keep the value in db user.idp as `idir` (and not update to `azureidir`) when user logs in with azure idir.(?)
eg: around line 157: 
```
if (newUser.idp === 'azureidir') newUser.idp = 'idir';
```
?

TimCsaky: but thinking more..
client app (eg bcbox) would need to also need to re-elvaluate user token.idp as always `idir`

app/src/components/constants.js
norrisng-bc: I think I threw in there as a "just in case", but yeah it's probably a bit redundant and unnecessary.

app/src/services/user.js
norrisng-bc: The implicit `azureidir` to `idir` conversion is handled at the auth middleware layer, so the UserService won't even see it. The user will just appear to it as `idir` (even though the underlying JWT field is actually `azureidir`), so it's as if there's no change or update needed on `user.idp` in the COMS DB.

norrisng-bc: But yes, the client does need to make this same assumption and handle it accordingly (e.g. if the JWT `identity_provider` is `azureidir`, and you want to search for IDP-level permissions the user has access to, make the COMS API call with `idp=idir` instead - don't just use that field from the JWT as-is).

...that is, unless we update the COMS-side code in a couple more places so that `azureidir` is silently converted to `idir` (e.g. wherever `idp` is accepted as a parameter), in addition to `currentUser`.

TimCsaky (approved):

